PR #253, opened 2015-01-14
More consistent spaces

hannesvdvreken:


GrahamCampbell:
We should actually be removing the spaces. Please follow PSR-2.

hannesvdvreken:
@GrahamCampbell check #254

PR #259, opened 2015-01-22
directly invoke callables instead of call_user_func.

staabm:
eases debugging, because of simpler stacktraces.

This is the only place where we have a strongly typed `callable` therefore this optimization cannot be made anywhere else.

### DIFF
--- a/src/Whoops/Handler/CallbackHandler.php
+++ b/src/Whoops/Handler/CallbackHandler.php
@@ -43,7 +43,10 @@ class CallbackHandler extends Handler
         $exception = $this->getException();
         $inspector = $this->getInspector();
         $run       = $this->getRun();
+        $callable  = $this->callable;
 
-        return call_user_func($this->callable, $exception, $inspector, $run);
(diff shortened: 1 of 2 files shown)


denis-sokolov:
As you can see from the Travis results, this does not exactly work.
PHP has two namespaces on a class: one for properties, another for methods.
You could try `$callable = $this->callable; $callable($exception, $inspector, $run)`;

staabm:
@denis-sokolov thx, had a slightly different patch locally and did some "trivial" changes where, which obviously broke it again. sorry for that.

with your suggested change, travis is happy again.

denis-sokolov:
I think it's a good idea to give your rationale in the comments right on that line.
It is easy to see that another developer in the future can easily simplify your code by removing a temporary variable and replacing it with a `call_user_func`. :)

Alternatively, even better, maybe there's a possibility for an automated test to check if there's a `call_user_func` in the trace?

staabm:
give me a few days, I will update that.

staabm:
@denis-sokolov here we go.

denis-sokolov:
Awesome!

PR #260, opened 2015-02-04
register the handleError() only for E_* levels defined in error_reportin...

staabm:
...g()

This improves performance, because handleError() is not invoked for every E_* level, but only those we are interessted in. This removed a lot of unnecessary invocations which lead to a no-op as the error-level will be filtered nevertheless.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -114,7 +114,7 @@ class Run
             class_exists("\\Whoops\\Exception\\Frame");
             class_exists("\\Whoops\\Exception\\Inspector");
 
-            set_error_handler(array($this, self::ERROR_HANDLER));
+            set_error_handler(array($this, self::ERROR_HANDLER), error_reporting());
             set_exception_handler(array($this, self::EXCEPTION_HANDLER));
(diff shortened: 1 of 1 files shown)


staabm:
I left `if ($level & error_reporting()) {` within `handleError` as the method is public and might get called by user-level-code

denis-sokolov:
This is a change in functionality. What if error reporting level changes mid-process?

staabm:
Is this a scenario Whoops is supposed to support? If so, I would expect to have a `Run->setErrorReporting($level);` method in the api instead of depending on the global error-reporting..

denis-sokolov:
I'm pretty sure we're supposed to support all regular error handling in PHP.
There is a lot of code changing error reporting level at runtime, including 3rd party library code that the user is unable to change to use our API. And what advantage would that bring anyway?

staabm:
Libs changing the error-reporting do it to hack arround things which would otherwise not be possible because of the way php works (and restore to prev behaviour right after that). IMO a regular app would define the error reporting at boostrap-time and not mid-process.

But I am fine with closing here, If you dont think so.

denis-sokolov:
I'd rather stay on the safe side here.

PR #263, opened 2015-02-17
Fix #262: Silenced errors won't fill error_get_last() array

### LINKED ISSUE #262: Silenced errors won't fill error_get_last() array
For example the following code (with warnings throwing exception) will not work as the error handler doesn't propagate error to the default handler:

    $success = @mkdir($dir, 0777, true);
    if (!$success) {
        $error = error_get_last();
        throw new RuntimeException("Creating directory '{$dir}' failed on error {$error['message']}");
    }

If directory creation fails, details on why it happened will not be there. To fix this, the error handler should return false on silenced errors.

So in Run::handleError() handler should return false if the error was silenced.

mahagr:
Pass silenced and ignored errors to default error handler.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -302,7 +302,8 @@ class Run
      * @param string $file
      * @param int    $line
      *
-     * @return bool|null
+     * @return bool
+     * @throws ErrorException
      */
     public function handleError($level, $message, $file = null, $line = null)
     {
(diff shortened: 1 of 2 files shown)


denis-sokolov:
Thank you for your contribution, great fix.
Would you please split the commit into two with the fix in one and the comment block fix in another?

mahagr:
Well, comment block fix is actually part of this fix as I changed the function always to return boolean.

denis-sokolov:
That's true, but not the `@throws` part.
Oh well, it's not a big change.
Thanks for the contribution.

PR #266, opened 2015-03-10
JsonResponseHandler pretty print

2bj:


GrahamCampbell:
:-1: This is broken on php 5.3. Just checked...

denis-sokolov:

As Graham mentioned, this breaks 5.3, which we support at the moment.

In addition, I assume that `JsonResponseHandler` consumers are machines and do not care about pretty printing. Developers probably use `PrettyPageHandler`.
(To be fair, I can't come up with a use case for `JsonResponseHandler` at all.)

tanuck:
Surely the use case is exactly what has been demonstrated? It's nice to have the JsonResponseHandler for debugging ajax requests in the console? You're not gonna want the entire HTML content from PrettyPageHandler in your console?

denis-sokolov:
I do want `PrettyPageHandler` in the console.
If you click on a `Preview` tab in your screenshot, you'll see the entire pretty HTML page.
(The scripts won't work, so you can't switch to other traceback levels, though)

Even if that's not an option, your alternative is `PlainTextHandler`. Why do you want it to be JSON?

Issue #267, opened 2015-03-12
error-code of \ErrorExceptions not handled properly?


staabm:
When using a error to exception "converter" like suggested on php.net [1], the errors'  `$errno` should go into the `severity` property.

Whoops instead uses the `code` property of the ErrorException instead [2].

Should we be able to handle both cases or should this be considered a "convention" whoops brings with it?

[1] http://php.net/ErrorException
[2] https://github.com/filp/whoops/blob/master/src/Whoops/Handler/PrettyPageHandler.php#L130-L134

denis-sokolov:
Good catch, @staabm!

That's really unfortunate that we have a bug like this.
I think we definitely will be setting `severity` to align with PHP.

The `level` though is more complicated.
At the moment the PHP model does not use the `level` property at all.
I do think that removing `level` in our code, as unlikely as it, might break applications that rely on `->getCode()` incorrectly. We shall keep it for the time being (that is, `getCode() === getSeverity()`), and mark a note somewhere that it's deprecated.

staabm:
> The level though is more complicated.

whats the `level` you are referring to?

denis-sokolov:
I meant `code`.

staabm:
ok, will provide a PR then

PR #268, opened 2015-03-13
error-code should be wrapped in ErrorException->severity

staabm:
WIP for issue https://github.com/filp/whoops/issues/267

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -130,7 +130,8 @@ class PrettyPageHandler extends Handler
         $code = $inspector->getException()->getCode();
 
         if ($inspector->getException() instanceof \ErrorException) {
-            $code = Misc::translateErrorCode($code);
+            // ErrorExceptions wrap the php-error types within the "severity" property
+            $code = Misc::translateErrorCode($inspector->getException()->getSeverity());
         }
 
         // List of variables that will be passed to the layout template.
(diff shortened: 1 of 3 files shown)


staabm:
@denis-sokolov here we go.

denis-sokolov:
Great job, awesome.
Do you mind squashing the commits together? The test and the change related to the test always belong together in my opinion.

staabm:
@denis-sokolov squashed the commits.

small side note: when you as a maintainer like squashed commits instead of separated you should IMO have scripts which would do that for you while merging.
otherwise correct and important PRs might not get merged just because they are contained of X commits instead of 1. Its like dont merge a PR because of CS nits.
But thats just my personal opinion, nothing you should be worried about.

thank you.

denis-sokolov:
I completely agree with you on a general idea of code style and pull requests.
I also do have tools to squash myself, git itself is plenty enough.

Rather, I thought, it was a matter of semantic meaning, and the author is better informed to make correct decisions. For instance, perhaps you see something that I didn't that made you choose a different decision. It seems I have been mistaken and in this particular case it was, indeed, completely arbitrary and I could have done that myself without any risk.

I have a procedure though, and, of course, were you to tell me that you can't or won't squash, I'd help you or pull what I want in whatever order I want myself.

Thank you for your contributions!

staabm:
@denis-sokolv thank for for your time and the great job you are doing.

Wasnt meant to be harsh.

PR #270, opened 2015-03-13
fixed ellipsis for non-utf8 apps

staabm:
When using Whoops in a non-utf8 app, the utf8 ellipsis char in Frame.php cannot be rendered.

With this PR we use the html-entity for this character instead which works in all encodings.

In a CP1252 App without this patch the stackframe view looks like http://imgur.com/FQOn54a


denis-sokolov:
Great fix!

_(And I feel sorry for people not using Unicode.)_

staabm:
>(And I feel sorry for people not using Unicode.). 

Thanks. My job is managing legacy apps. A lot of them and none with a proper charset :-/

Issue #272, opened 2015-03-18
PrettyPage: Keyboard Navigation


staabm:
To enhance the user experince on pretty-pages I would like to propose adding the following keypress handlers:

- Arrow Up/Down to navigate between stack frames (the column on the left, which is ATM only discoverable using the mouse)
- Page Up/Down to scroll the detail panel on the right

Do you miss anything else which might be a good thing to enhance the UX?

Atm you can scroll the frame which has the focus, so either scroll stackframes window (without showing frame details like a click would) or scroll the details window. I don't like using the mouse in between therefore I suggest the above enhancements.

As an alternative we could scroll the stackframes using Arrow Left/Right instead which would preserve browsers default behavior (so you could still scroll the "active window" depending on where you clicked before)

denis-sokolov:
I'd like to minimize the intrusion into default controls. We should focus the main panel on load to allow for regular scrolling.

Then we need to solve the frame navigation. Left and right though is not particularly straightforward.
Perhaps `Ctrl+↑` and `Ctrl+↓`? We can show a tiny hint in the UI (thought less prominent that in my mockup):
![1](https://cloud.githubusercontent.com/assets/113721/6764466/3a831356-cfb9-11e4-8d24-190f05460be9.png)

staabm:
Will provide a PR next week

PR #273, opened 2015-03-25
added inline-help for PrettyPage Handler

staabm:
added a inline-help overlay which show up when the user hovers the "HELP" button at the top right.
it describes what the user sees within woops and also mentions possbile keyboard shortcuts (to be implemented in #272)

264zlqd

If we agree on this help-overlay and its styling etc. I would goon and implement the keyboard shortcuts.


denis-sokolov:
I can stand behind the general idea.

denis-sokolov:
Things I'd love to see improved:
* Avoid marketing language: "at a glance", "integration", "immediately see", "click the button". Currently the language is also a bit inconsistent. "see the line throwing", but "shows all information". I'd suggest short, and formal "Code snippet where the error was thrown", "Server state information", "Exception message and its type".
* Note that there may be more information in the big area on the bottom right. There can be optional data if the application throwing it provides. There's also request data in there.
* Perhaps the hints don't need a title, and can consist only of a short message?

Regardless of my suggestions, it's a good idea and good work is done already, good job!

staabm:
Updated per comments. WDYT?

inlinehelp-v2

denis-sokolov:
I think this looks good!

staabm:
Ok, will add the js handlers for ctrl+up/down then

staabm:
finally added the keyboard shortcuts. this fixes https://github.com/filp/whoops/issues/272

staabm:
@denis-sokolov ready to land.

denis-sokolov:
Do we already have Ctrl+arrows navigation working?

staabm:
Jepp, tested it

denis-sokolov:
Oh, I see, as part of this PR. That got me confused. I'll take a look at it, but great job!

staabm:
@denis-sokolov thanks for merging. would you mind tagging another release?

denis-sokolov:
@staabm, done.

PR #274, opened 2015-03-26
fix typo in API Documentation

matthew-james:
should be singular - https://github.com/filp/whoops/blob/master/src/Whoops/Handler/PrettyPageHandler.php#L377


denis-sokolov:
Thanks!

PR #275, opened 2015-03-30
HHVM fix - flush before exit

colin-kiegel:
Whoops doesn't work for me with HHVM 3.6 due to a minor incompatibility between PHP and HHVM.

See facebook/hhvm#4055. The solution is to add flush() before exit(). IMO this workaround can be removed as soon as HHVM fixes this bug upstream in all their LTS releases.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -285,6 +285,7 @@ class Run
         }
 
         if ($willQuit) {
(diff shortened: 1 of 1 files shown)


denis-sokolov:
That's cool, thank you.

colin-kiegel:
Thx. It was a bit tough to spot - but at least that got me xdebug running with HHVM now! Yess! Bugs, here I come! ;-)

PR #277, opened 2015-04-10
Fixed help overlay flicker

JayBizzle:
When hovering over the help button, if you move your mouse pointer slowly, the overlay flickers because the button loses 'focus' when the overlay is displayed.

The added CSS property will prevent the overlay from receiving pointer events. It isn't a perfect solution though. It works in all modern browsers, but no versions of IE less than IE11.

http://caniuse.com/#feat=pointer-events


denis-sokolov:
Thanks. I'd rather have a better solution for help showing up overall, but this is a nice fixup for the time being.

PR #279, opened 2015-04-24
Display help button above help overlay

adduc:
This helps with flickering when the cursor moves over the help button in Chrome (possibly other browsers).


adduc:
Didn't see #277, closing as this has already been resolved.

PR #280, opened 2015-05-08
Saving some memory for echo.

dseguy:


GrahamCampbell:
Does this actually make any noticeable difference?

denis-sokolov:
Thank you for your contribution.
We will not be changing this pattern of usage for lack of reasons to switch. But please, continue submitting ideas in the future!

PR #282, opened 2015-05-09
Specific-fix for some protocols that quit the page...

Geolim4:
...Such as IntelliJ Platform REST API.

Indeed, as it returns a valid http response, the browser quits the page and then leaves the Whoops page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ vendor/*
 report/*
 phpunit.xml
 composer.lock
+/.idea/
--- a/docs/API Documentation.md
+++ b/docs/API Documentation.md
@@ -365,7 +365,7 @@ PrettyPageHandler::addResourcePath(string $resourcesPath)
 
 // Sets an editor to use to open referenced files, either by
 // a string identifier, or as an arbitrary callable that returns
-// a string that can be used as an href attribute.
+// a string or an array that can be used as an href attribute.
 // Available built-in editors are:
 // - sublime
 // - emacs
@@ -373,6 +373,15 @@ PrettyPageHandler::addResourcePath(string $resourcesPath)
 // - macvim
 PrettyPageHandler::setEditor(string $editor)
 PrettyPageHandler::setEditor(function ($file, $line) { return string })
+
+// Additionally you may want that the link acts as an ajax request (e.g. Intellij platform)
+PrettyPageHandler::setEditor(function ($file, $line) { 
+        return array(
+            'url' => "http://localhost:63342/api/file/?file=$file&line=$line",
+            'ajax' => true
+        ); 
+    }
+)
  #=> null
 
 // Similar to PrettyPageHandler::setEditor, but allows you
@@ -387,6 +396,8 @@ PrettyPageHandler::setEditor(function ($file, $line) { return string })
 // $handler->addEditor('whatevs', 'whatevs://open?file=file://%file&line=%line')
 PrettyPageHandler::addEditor(string $editor, $resolver)
  #=> null
+ 
+ 
 
 PrettyPageHandler::handle()
  #=> int | null
--- a/docs/Open Files In An Editor.md
+++ b/docs/Open Files In An Editor.md
@@ -30,7 +30,7 @@ $handler->setEditor(function($file, $line) {
 
 ```
 
-You can add PhpStorm support with [PhpStormOpener](https://github.com/pinepain/PhpStormOpener#phpstormopener) (Mac OS X only):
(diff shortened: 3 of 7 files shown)


denis-sokolov:
Hm, interesting.
I am worried that on the flip side an Ajax request might not trigger other editors to open.
Could you try a simpler `target=_blank` instead of preventing the default and see if that solves your problem?

Geolim4:
You right, but the problem is as it return a valid http content body, this wont close the tabs but let it open, but i have another idea: What about giving possibility to return an array of attribute with getEditorHref() ?

Current behaviour: Return new link href. (as string)

Proposed behaviour: Return new link href (as string) OR return an array of attributes, e.g. [href => 'blahh', onclick => 'blahh', title => 'blahh'].

Look better for you? :)

denis-sokolov:
No, I dislike giving such power, we'd need to maintain that.
Can you clarify, what is your issue about? If you set your custom links to something custom like `intellij://`, the browser closes the tab after you press on the link?
What about if you set it to `mailto:`? Is it specific to your browser, perhaps? Try another browser.
I'd like to know exactly the circumstances causing this to design the best solution.

Geolim4:
In fact this is not an intellij "protocol" but mostly a local host with a custom port which allow you to send a command to the IDE for opening the file, e.g.
http://localhost:63344/api/file/?file=path_to_the_file&line=line_number&focused=true

The problem is that the API send a valid http body, which cause the browser leaving the page. 
This behavior is totally normal and is the same with all browser (ff, chrome, ie, safari, opera etc.).

staabm:
Why dont you adjust the api?

denis-sokolov:
@staabm, there's not much the server can do to control the browser in that situation, I believe. At best you send 204, but browser behavior is not strictly defined.

@Geolim4, I would accept a patch that would allow callable addEditor/setEditor parameters to return an array: `['url' => '...', 'ajax' => true]`.

Geolim4:
@staabm In fact i have no control to the API, as it handled & managed by phpStorm himself (internally). Also i would like to keep Whoops vanilla as possible.

@denis-sokolov Great decision, i will modify the Pr in hours. Thanks.

Geolim4:
@denis-sokolov
Also i saw in the doc that there's a PhpStormOpener support, but as you can see here:
https://youtrack.jetbrains.com/issue/IDEA-65879#comment=27-913958
JetBrain Team announced the RestFull API (that i am currently trying to implement).
Doc availlable here: http://develar.org/idea-rest-api/

This should deprecate the current (Mac OSX only) phpStorm support, did you confirm it ?

denis-sokolov:
Curent phpStorm support? I [don't think we have it](https://github.com/filp/whoops/blob/f924458af1b06fe6e03fdb0c29322ba011c90ce7/src/Whoops/Handler/PrettyPageHandler.php#L70).

Geolim4:
What About this?
https://github.com/filp/whoops/blob/9f451fbc7b8cad5e71300672c340c28c6bec09ff/docs/Open%20Files%20In%20An%20Editor.md

denis-sokolov:
Ah, well, yes. But it's just a recipe for what you can do in your code. This is unrelated to our code.

Geolim4:
This is why i rewritten this part in @ccd137e :)
All is well else ?

denis-sokolov:
Looks very good! I especially like the proper usage of a data attribute in HTML as opposed to a contrived `onclick` or something.

One final thing - please don't change `getEditorHref` to return an array. Even with options, `getEditorHref` should only return a string, a URL. I suppose you will need a new boolean returning `getEditorAjax` to retrieve the setting in the template.

Geolim4:
So we can do a special method like `getEditorAjax()` ?

denis-sokolov:
Yes.

Geolim4:
Does it look better now? :)

denis-sokolov:
It does. It's not perfect yet, though.
Your implementation of `getEditorAjax` [throws](https://github.com/filp/whoops/pull/282/files#diff-cb7af2a30dcc0c802f01e3dd9c020ef5R374) if the user does not return an array or if the array does not contain `ajax` key. But the majority of users won't, and you should silently handle this.
Consider a helping private function that `getEditorHtml` and `getEditorAjax` share. The function would take whatever the current editor is and normalize it into the widest possible format, with all options defined: `['url' => "", 'ajax' => false]`. Then `getEditorAjax` and `getEditorHtml` are simply `return $this->myHelper(...)['key']` and no matter how many options we add or modify in the future, we only need to edit that helper.

Geolim4:
I will modify that tonight, rhx :)

Geolim4:
I applied an update, it's all ok now?

Issue #256, opened 2015-01-19
Hide the error page


Petah:
It would be nice if there was like a close button that removed or hide the whoops page. 

This would be handy in the case that there was some output flushed before an error, whoops will sit on top of it and it cannot be seen.

denis-sokolov:
An interesting idea, but I don't think it warrants a UI space.
PrettyPageHandler is supposed to only be enabled in development, and there is no valuable information that could be outputted. And if by chance you really need to see what's there, just open developer tools and hide or delete the Whoops container DOM element.

GRMule:
I came here looking for this exact thing.

As you said, the use-case is development. It is entirely plausible, likely even, that there is some debugging output behind the Whoops error message that would be useful in resolving the error that Whoops is reporting but doesn't make it into the formatted output. 

You're generally not wrong that it isn't a big deal to open dev tools and wipe the div out, *but*, that also takes time. If I don't already have the console loaded, or I'm working with a device without a convenient console... at any rate, I don't personally see why **not** have a fast and easy way to hide the output without extra steps.

I implemented it by adding an absolute positioned div with an X in it to the top/right of the header, so it didn't really take any additional UI space at all. A tiny addition to the .js file, and it's done. I created a fork and implemented these changes. Even though this bug was declined, I sent a pull request (PR #283) anyway because why not. No offense if you don't take it :)

denis-sokolov:
@staabm, @GrahamCampbell, could you please provide your input?
I believe this is a barely useful feature and adding the UI for "quitting" the error is causing more confusion than benefit. What do you guys think?

Petah:
A hotkey then? Better than opening up the debugger and deleting the element.

denis-sokolov:
A hotkey alone is a cop out. Acceptable, but rather points to a problem not solved. Like a checkbox in advanced options. But this got me thinking. It's not a "hide the error page" feature. It's a "show me the raw output of my PHP script" feature. Instead of "X" we need something akin to "view raw script output", and instead of "+" we need "return to Whoops". Then it's entirely clear and is much easier to accept.
Thoughts?

Petah:
Sounds good!

GRMule:
[...]

PR #283, opened 2015-05-19
Added a show/hide toggle for Whoops output

GRMule:
Miscellaneous debugging output is covered by Whoops. This toggle allows such background output to easily be viewed without relying on developer tools to remove/hide the Whoops container, related to issue #256 

Adds an fixed-position X to the upper right corner. Clicking this will hide the top-level Whoops container, and the element will show a + instead. A second click will restore the Whoops container visibility.

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -11,23 +11,6 @@ body {
     text-decoration: none;
   }
 
-.Whoops-hider {
-  position: fixed;
-  top: 0px;
-  right: 20px;
-  font-size: 14px;
-  color: red;
-  background-color: #22222;
-  padding: 6px;
-  cursor: pointer;
-  font-weight: bold;
-  z-index: 100;
-}
-.Whoops-hider:hover{
-  background-color: red;
-  color: white;
-}
-
 .container{
     height: 100%;
     width: 100%;
@@ -36,6 +19,9 @@ body {
     padding: 0;
     left: 0;
     top: 0;
+}
+.container.hidden {
+    display:none;
 }
 
 .branding {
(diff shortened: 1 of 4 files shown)


staabm:
In case @denis-sokolov accepts this feature, it should also be promoted using the HELP overlay

GRMule:
@staabm I added the help layer text, good input!

src/Whoops/Resources/js/whoops.base.js
```diff
@@ -50,13 +50,25 @@ Zepto(function($) {
...
 	  clip.on("load", function() {
 		  $clipEl.show();
+		  $clipHelpEl.show();
 	  });
   }
-  
+
+  $(document).on('click', '.Whoops-hider', function () {
+	  var $container = $('.Whoops.container');
+	  if ($container.hasClass('hidden') === false) {
```
denis-sokolov: Or just `if (!$container.hasClass('hidden'))`.

src/Whoops/Resources/js/whoops.base.js
```diff
@@ -50,13 +50,25 @@ Zepto(function($) {
...
+		  $container.hide().addClass('hidden');
```
denis-sokolov: It's better to hide the element with CSS using `.hidden { display: none }`. Even better is to use an [attribute `hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) for that. (Likely polyfill it as `[hidden] { display: none; }`)

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -379,9 +397,9 @@ pre.prettyprint, code.prettyprint {
 }
 
 #help-exc-message {
-	position: absolute;
-	left: 65%;
-	top: 10px;
+    position: absolute;
+    left: 52%;
```
denis-sokolov: Is there a specific reason for moving the exception message to the left?

denis-sokolov:
The clipboard improvement is great, but it has nothing to do with this pull request.
One of the reasons why it's great to commit independently is I'd love to merge the clipboard fix right away, but I'm not ready to merge the main PR yet.

denis-sokolov:
For this to be merged, the discussion in #256 needs to be resolved.

src/Whoops/Resources/css/whoops.base.css
GRMule: That's moving the exception *help* to the left, to make room for the hider help message. Originally, there was no room (on my screen, anyway, ~1260px max width) for the hider message in the upper corner. On a smaller screen (ex. 1024 MW), they juuuust approach overlapping, because the exception help is using a percentage for its position.

src/Whoops/Resources/js/whoops.base.js
GRMule: My personal code style dictates expressive conditionals, I use full comparisons rather than the negation operator as a matter of practice and habit. I've changed it to use the ! operator, per your input.

src/Whoops/Resources/js/whoops.base.js
GRMule: I added the .container.hidden rule, per your input.

I don't agree with you about the hidden attribute, "This attribute must not be used to hide content that could legitimately be shown." :)

GRMule:
Apologies for the combined commit. This is my first PR, next time I will not piggy-back changes, I see now that it makes selective adoption difficult. Should this PR wind up rejected, I will re-submit the clipboard change independently.

src/Whoops/Resources/css/whoops.base.css
denis-sokolov: Ooh, right, thanks.

src/Whoops/Resources/js/whoops.base.js
denis-sokolov: I don't feel strongly about it, but thanks.

src/Whoops/Resources/js/whoops.base.js
denis-sokolov: Support fixing is trivial, as I mentioned: `[hidden] { display: none; }`.
But your reading of the spec has taught me something new, thank you.

denis-sokolov:
Again, sorry for the delay.
Given the scope of changes in #292, I will ask you to wait a bit more until that's done.
Once that's done, we'll fit this into the new design and merge it in!
Please speak up if this gets lost again!

denis-sokolov:
@GRMule, #292 has been merged in, thus your feature can now be added as well!
Do you maybe want to update your PR to resolve the conflicts and test the new looks?

denis-sokolov:
Cleaning up pending prs.
Whenever there's any update, will reopen no problem.

PR #284, opened 2015-05-20
PHP7 and EngineException handling

adduc:
PHP7 introduces a few new classes related to Exception ([RFC](https://wiki.php.net/rfc/engine_exceptions)). BaseException is created as a parent of Exception, and EngineException is created as a child of BaseException. As a result, the exception handler may receive objects that do not match Exception (EngineException is not an instance of Exception).

If we were to aim for PHP 7 compatibility only, I'd suggest changing all typehints in Whoops to BaseException, to allow for both Exceptions and EngineExceptions to be passed throughout.

If we were to aim for PHP 5 and PHP 7 compatibility, as suggested by @LawnGnome, we would need to get rid of the type hint in the exception handler ([Run::handleException](https://github.com/filp/whoops/blob/c982fe62c44798c433229cb0425c61b487cc1883/src/Whoops/Run.php#L233)). We would also need to remove all other typehints for Exception, as they would also need to handle both Exception and EngineException.

I'm opening this pull request to start discussion, I don't intend for the changes within to be merged without further refinement (still working on a unit test to test catching EngineException).


staabm:
We should not land this before php7 was released as an alpha or rf, because there are several open PR which might change the Exception-hierarchy or the names of those classes.

src/Whoops/Exception/Inspector.php
```diff
@@ -28,7 +28,7 @@ class Inspector
     /**
      * @param Exception $exception The exception to inspect
      */
-    public function __construct(Exception $exception)
```
GrahamCampbell: You need to update the phpdoc.

src/Whoops/Exception/Inspector.php
```diff
@@ -131,7 +131,7 @@ public function getFrames()
      * @param  Exception $exception
      * @return array
      */
-    protected function getFrameFromException(Exception $exception)
+    protected function getFrameFromException($exception)
```
GrahamCampbell: You need to update the phpdoc.

src/Whoops/Handler/Handler.php
```diff
@@ -74,7 +74,7 @@ protected function getInspector()
     /**
      * @param Exception $exception
      */
-    public function setException(Exception $exception)
+    public function setException($exception)
```
GrahamCampbell: You need to update the phpdoc.

src/Whoops/Handler/HandlerInterface.php
```diff
@@ -27,7 +27,7 @@ public function setRun(Run $run);
      * @param  Exception $exception
      * @return void
      */
-    public function setException(Exception $exception);
+    public function setException($exception);
```
GrahamCampbell: You need to update the phpdoc.

src/Whoops/Run.php
```diff
@@ -95,7 +95,7 @@ public function clearHandlers()
      * @param  Exception $exception
      * @return Inspector
      */
-    protected function getInspector(Exception $exception)
+    protected function getInspector($exception)
```
GrahamCampbell: You need to update the phpdoc.

src/Whoops/Run.php
```diff
@@ -230,7 +230,7 @@ public function writeToOutput($send = null)
      * @param  Exception $exception
      * @return string    Output generated by handlers
      */
-    public function handleException(Exception $exception)
+    public function handleException($exception)
```
GrahamCampbell: and in all futher places too...

adduc: Yes... I did this hastily to get Whoops working in PHP 7 as I test other code bases, but I suppose that's a poor excuse.

denis-sokolov:
Thank you, @adduc, for a very important topic raised and a few solutions proposed.
At the moment I agree with @staabm, I think we need to wait for some level of PHP 7 stability to not redo our changes five times over.
But let's keep this PR alive and open for discussion.

GrahamCampbell:
> some level of PHP 7 stability

PHP 7.0 is already locked for RFCs isn't it?

staabm:
> PHP 7.0 is already locked for RFCs isn't it?

yes it is, but they still discuss this change on the mailing list and in a few PRs
https://github.com/php/php-src/pull/1274
https://github.com/php/php-src/pull/1282
https://github.com/php/php-src/pull/1284

GrahamCampbell:
Does anyone know when the first alpha release is planned? It can't be too far off can it?

staabm:
@GrahamCampbell see the timeline https://wiki.php.net/rfc/php7timeline
It seems we can expect a RC version Mid-June

GrahamCampbell:
June says RC. That's not alpha?

GrahamCampbell:
alpha -> beta -> RC -> stable

GrahamCampbell:
or are they using RC to describe all pre-stable releases?

staabm:
Right, JUNE says RC. Since I expect at least 5-8 beta+alpha releases I think they are not on schedule.
We should at least wait for the 2nd or 3rd alpha.

GrahamCampbell:
>  I think they are not on schedule.

lol - I suspected that would happen. PHP 7 is a lot of work in a short time span.

GRMule:
Last I saw (earlier today), they're talking about October.

GrahamCampbell:
What about other types of exception too?

GrahamCampbell:
"ParseException" doesn't extend "Exception" or "EngineException" in php 7.

GrahamCampbell:
https://speakerdeck.com/player/46ff98632f8c4b92b0536f7a68e0e794 is useful. The exception stuff is in the middle.

staabm:
Soo the final PR landed: https://github.com/php/php-src/pull/1284

staabm:
@denis-sokolov looks like this is ready to land?

src/Whoops/Exception/Inspector.php
```diff
@@ -26,15 +26,15 @@ class Inspector
     private $previousExceptionInspector;
 
     /**
-     * @param Exception $exception The exception to inspect
+     * @param Throwable $exception The exception to inspect
```
GrahamCampbell: You need a leading slash here

src/Whoops/Exception/Inspector.php
```diff
@@ -26,15 +26,15 @@ class Inspector
...
      */
-    public function __construct(Exception $exception)
```
GrahamCampbell: You need to remove Exception from the imports everywhere

Petah:
Would be good to get this merged as we are beta testing PHP 7 now.

GrahamCampbell:
Removing typehints from public methods is a breaking change and require a major version bunp.

GrahamCampbell:
> Would be good to get this merged as we are beta testing PHP 7 now.

There are still issues in this PR that haven't been addressed yet.

src/Whoops/Exception/Inspector.php
staabm: to be precise: backslash ;)

adduc:
@GrahamCampbell I'm sorry I haven't been more responsive, I've updated to remove Exception from imports in applicable files, as well as corrected the typehint used in documentation to correctly resolve to the global namespace.

tests/Whoops/Exception/InspectorTest.php
```diff
@@ -12,21 +12,21 @@
...
-     * @param  string    $message
-     * @param  int       $code
-     * @param  Exception $previous
+     * @param  string     $message
+     * @param  int        $code
+     * @param  \Throwable $previous
      * @return Exception
      */
-    protected function getException($message = null, $code = 0, Exception $previous = null)
+    protected function getException($message = null, $code = 0, $previous = null)
```
GrahamCampbell: This is wrong?

GrahamCampbell: The previous exception really does have to be an exception.

denis-sokolov:
> Removing typehints from public methods is a breaking change and require a major version bunp.

@GrahamCampbell, can you clarify, how? In other words, do you have an example of a code that could break after removing a typehint?

denis-sokolov:
As for merging this quicker, I can't find any official announcement that the current solution is set is stone and will not change. I will not release a Whoops version with support for that what might still change.

If there is a considerable need to use Whoops on PHP 7 before anything is clear about the final design, we may create a side-branch and release that on Packagist. But is there really such a need?

Petah:
I think there should at least be a dev branch available through composer as
currently I have to disable whoops to beta test php7

denis-sokolov:
Can you not do that automatically?
```php
if (version_compare(PHP_VERSION, '6.0.0', '<')) {
    $whoops->register();
}
```

Petah:
The point is I don't want to disable whoops error handling, I want it, but it fails in some cases in PHP 7.

denis-sokolov:
Fair enough.
Well, I've now provided the [`php7` branch](https://packagist.org/packages/filp/whoops#dev-php7) on the official registry that includes @adduc changes, feel free to use that.

staabm:
> @GrahamCampbell, can you clarify, how? In other words, do you have an example of a code that could break after removing a typehint?

@denis-sokolov A method overriding the method which was changed but still uses the typehint will emit a E_STRICT

http://3v4l.org/3Om6J

denis-sokolov:
Thank you.

Interesting. Means we might need to do more magic workarounds to have a codebase that works across versions. Not that I think extending `Whoops` class is particularly worthy. I would even mark it as `final` if it was up to me.

staabm:
Agree that marking `whoops` final is a good idea.

GrahamCampbell:
Removing typehints of methods on non-final classes is a breaking change so requires a semver MAJOR version bump.

Petah:
So is there any plans to get php7 support in stable? Its approx 2 weeks to slated release.

hultberg:
For PHP 5 and PHP 7 compability one could check if class `EnigeException` exists and if not just create it as a fallback and let it extend `Exception` and that would make it catch `Exception` in PHP 5 and `EngineException` in PHP 7.

denis-sokolov:
Everybody, please check #312 and discuss.

denis-sokolov:
With our call to move to a BC break and Whoops 2.0, I am back to this from #312 

@adduc, do you want to update your PR to resolve the new conflicts?

mahagr:
@denis-sokolov As @adduc seems to be silent, you can easily merge this yourself.

Make a new branch from the `master`, pull the changes from this pull request (=fork), resolve conflicts and push back to master. This pull request should be automatically closed by that action.

denis-sokolov:
Thanks, @mahagr. I know. I will have to if I don't get an updated PR until the end of the week.
We really need to release soon.

staabm:
@mahagr feel free to pick the changes and provide a conflict resolved Pr.

mahagr:
There are no conflicts, pull was successful without any manual work. But for some reason GitHub still thinks that there are conflicts. Someone with commit rights to upstream master needs to merge this manually.

mahagr:
Maybe the conflicts are from the fact that there's local php7 branch?

moay:
Still not fixed with alpha1. I get this error when producing an error:

	Whoops\Exception\ErrorException thrown with message "Uncaught TypeError: Argument 1 passed to Whoops\Exception\Inspector::getTrace() must be an instance of Exception, instance of Error given, called in /some/path/vendor/filp/whoops/src/Whoops/Exception/Inspector.php on line 95 and defined in /.../vendor/filp/whoops/src/Whoops/Exception/Inspector.php:163
	Stack trace:
	#0 /some/path/vendor/filp/whoops/src/Whoops/Exception/Inspector.php(95): Whoops\Exception\Inspector->getTrace(Object(Error))
	#1 /some/path/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php(131): Whoops\Exception\Inspector->getFrames()
	#2 /some/path/vendor/filp/whoops/src/Whoops/Run.php(256): Whoops\Handler\PrettyPageHandler->handle(Object(Error))
	#3 [internal function]: Whoops\Run->handleException(Object(Error))
	#4 {main}
	  thrown"

	Stacktrace:
	#0 Whoops\Exception\ErrorException in /some/path/vendor/filp/whoops/src/Whoops/Exception/Inspector.php:163

`getTrace()` still expects an `\Exception`, so `\ErrorException`s won't work.

flaviocopes:
@moay fixed in the repo by https://github.com/filp/whoops/pull/352, not yet released

PR #285, opened 2015-05-28
getEditorHref returns false if editor is not set

andrejguran:
When editor is not set, getEditor returns false and getEditorHref throws the UnexpectedValueException exception.


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -324,6 +324,10 @@ public function getEditorHref($filePath, $line)
     {
         $editor = $this->getEditor($filePath, $line);
 
+        if (!$editor) {
+            return false;
```
staabm: needs a `@return string|boolean` then above

denis-sokolov:
Right, good one, thanks.

PR #288, opened 2015-05-29
Removed broken badge

GrahamCampbell:


denis-sokolov:
You should have mentioned the reason. Depending.in seems to be dying, as the invalid SSL certificate has not been fixed [since December](https://github.com/toopay/depending/issues/17).

GrahamCampbell:
> You should have mentioned the reason. Depending.in seems to be dying

I didn't know. I always thought the badge was pointless anyway. ;)

PR #290, opened 2015-06-05
Move $_SERVER superglobals to the end of the globals-table.

### LINKED ISSUE #209: PrettyPageHandler: Order of superglobals in detail view
I think we should re-order the superglobals in the detail view of the PrettyPageHandler.

I would expect to have them sorted starting from the request related ones to the more global ones. E.g. the $_SERVER array won't change often between requests and therefore the possibility that it holds information which is related to a particular request-error is rather low. 

IMO the GET/POST/FILES should be rendered first, then Cookies+Session and at the very least SERVER and ENV.

What do you think?

staabm:
closes #209.


GrahamCampbell:
:+1:

PR #301, opened 2015-07-22
Added JSON_PARTIAL_OUTPUT_ON_ERROR to json_encode to handle recursive stack traces

oligriffiths:
If the stack trace contains recursion, the JSON output is null by default. The JSON_PARTIAL_OUTPUT_ON_ERROR flag sets the recursive node to null thus allowing the trace to be rendered.

note that JSON_PARTIAL_OUTPUT_ON_ERROR is only available on PHP 5.4+, so support is added for that

### DIFF
--- a/src/Whoops/Handler/JsonResponseHandler.php
+++ b/src/Whoops/Handler/JsonResponseHandler.php
@@ -84,7 +84,8 @@ class JsonResponseHandler extends Handler
             header('Content-Type: application/json');
         }
 
-        echo json_encode($response);
+        echo json_encode($response, defined('JSON_PARTIAL_OUTPUT_ON_ERROR') ? JSON_PARTIAL_OUTPUT_ON_ERROR : 0);
+
         return Handler::QUIT;
     }
 }


src/Whoops/Handler/JsonResponseHandler.php
```diff
@@ -84,7 +84,12 @@ public function handle()
             header('Content-Type: application/json');
         }
 
-        echo json_encode($response);
+        if (version_compare(phpversion(), '5.4', '<')) {
```
GrahamCampbell: Faster to use the `PHP_VERSION` constant.

GrahamCampbell: Infact, it would be MUCH better to just check if `JSON_PARTIAL_OUTPUT_ON_ERROR` is defined.

GrahamCampbell: Just write:

```
echo json_encode($response, defined(JSON_PARTIAL_OUTPUT_ON_ERROR) ? JSON_PARTIAL_OUTPUT_ON_ERROR : 0);
```

oligriffiths: Good shout.

denis-sokolov:
Did you notice that the constant is actually only available in PHP 5.5, not 5.4?
This is another good reason why you should check the existence of a feature (constant), not a version number. Would you be willing to update your PR?

oligriffiths:
Done. Misread the docs about the version

GrahamCampbell:
Should be squashed to one commit.

oligriffiths:
Done

denis-sokolov:
This is great, thank you!

oligriffiths:
Any time, great package!

PR #304, opened 2015-08-11
Bugfix/fix issue with stacktrace

Vmak11:


denis-sokolov:
Thank you for your contribution. This seems a useful addition.

src/Whoops/Run.php
```diff
@@ -395,7 +395,7 @@ private function writeToOutputNow($output)
         return $this;
     }
 
-    private static function isLevelFatal($level)
+    public static function isLevelFatal($level)
```
GrahamCampbell: This feels dodgie to me.

GrahamCampbell: Maybe we should move this method somewhere.

Vmak11: Yes, agree, how about Whoops/Util/Misc.php?

PR #305, opened 2015-08-12
Hotfix/move is level fatal

Vmak11:


src/Whoops/Exception/Inspector.php
```diff
@@ -142,22 +143,7 @@ protected function getTrace(\Exception $e)
...
-                $fatal = true;
-                break;
-
-            default:
-                $fatal = false;
-                break;
-        }
-
-        if (!$fatal) {
+        if (!Misc::isLevelFatal($e->getCode())) {
```
staabm: Should be `e->getSeverity();`?

Vmak11: Agreed, sorry I was thinking about a regular exception and not an error exception but we return those on line 143 already so we can change it back to getSeverity

staabm: would be great if you could add a unit test for this case.

src/Whoops/Util/Misc.php
```diff
@@ -41,4 +41,21 @@ public static function translateErrorCode($error_code)
         }
         return "E_UNKNOWN";
     }
+    
+    /**
+     * Determine if an error level is fatal (halts execution)
+     * 
+     * @param int $level
+     * @return boolean
```
GrahamCampbell: bool, not boolean

denis-sokolov: [Can be any](http://www.phpdoc.org/docs/latest/references/phpdoc/types.html#abnf), though.

GrahamCampbell: `bool` is the correct type, and nothing else

GrahamCampbell: see php's docs, not some docs for some unoffical library

denis-sokolov: How does PHP documentation relate to this, though?
This is not PHP code, this is a documentation string, known as PHPDoc.
Nowadays the primary tool to generate documentation from PHPDoc is phpDocumentor, and it has a fairly extensive documentation on how exactly PHPDoc is supposed to look like, making this the closest thing to a standard we had for a long time.
Now there's [a PSR proposal](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#abnf) for the standard, authored by the same people from phpDocumentor.
How is that a "some unofficial library"?

Now, the proposed standard does not actually allow `boolean`, and it's a good idea to use `bool`, but it does not change the fact, that today by all the best guesses, `boolean` is as valid phpdoc as `bool` is.
Not to mention, that even after PSR proposal is accepted, all tools will always support `boolean` for backwards compatibility anyway.

GrahamCampbell: > Now, the proposed standard does not actually allow boolean

Yes it does. It defines as an alias, identical to what php does.

denis-sokolov: > Yes it does. It defines as an alias, identical to what php does.

Great if it does, then your correction about `bool` is even less useful.
But actually I can't find it in [the proposal](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#abnf). Searching the entire document I find 0 instances of "boolean".

GrahamCampbell: You can use any alias as defined in php.

PR #308, opened 2015-09-27
table cell renderer

kminek:
this PR allows user to set custom table cell renderer - i can for example use symfony/var-dumper:

```php
$handler->setCellRenderer(function ($value) {
    ob_start();
    dump($value);
    return ob_get_clean();
});
```

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -77,6 +77,12 @@ class PrettyPageHandler extends Handler
     );
 
     /**
+     * Renders HTML for table cell
+     * @var callable
+     */
+    private $cellRenderer;
+
+    /**
      * Constructor.
      */
     public function __construct()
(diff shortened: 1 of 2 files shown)


denis-sokolov:
I'm not sure whether we need to open this to customization. We should just bundle the symfony var dumper!

kminek:
not sure if it should be bundled in composer.json. it's presence could be detected and it could be used instead of print_r with some additional styling to match current layout :)

staabm:
I agree that we should simple built the var_dumper into whoops.

denis-sokolov:
@kminek, why not, what is the benefit of not using it always? It's more useful and more aesthetically pleasing than the standard output, right?

kminek:
if you don't mind having first dependency - then yes, it should be bundled :) i'm totally positive with that :) i just noticed that whoops has no dependencies so far so i though the "philosophy" was to stay with no dependencies :)

denis-sokolov:
Hm. I never thought of that. I don't see any specific benefit of having no dependencies. The selling point says "Stand-alone library with (currently) no required dependencies", so I'm not sure.

@filp, do you have an opinion on why is it beneficial to have no dependencies and whether you want the project to continue that way?

filp:
@denis-sokolov Having dependencies isn't really an issue for me. Not having any dependency maybe sounds good in terms of marketing whoops, but it really makes very little difference.

Just have to remember to update the readme to remove the line about not having any dependencies :wink:

staabm:
Easiest thing would be "use symfony var-dumper when available" and add it as a suggested dependecy. That way we would still be a standalone lib and poeple can opt-in to the feature

filp:
@staabm :+1:

denis-sokolov:
@kminek, or anyone else willing, please consider updating the pull request with the proposed flow of a suggested dependency.

staabm:
@denis-sokolov here we go 320

denis-sokolov:
@kminek, to conclude, thank you for your contribution.
As for this topic, please bring in your comments and suggestions on #321.
As I said, we will not be allowing customization.

PR #309, opened 2015-09-29
Silex 2.0

pawook:
Here I changed provider for a new version of Silex - 2.0.
I think it should be on a new branch while Silex 2.0 is under development, but I think I cannot create new branch in this project.
Am I doing right making a pull request, or what is the right way to do such things?
Sorry if my question sounds silly, I am pretty new to Github :)

### DIFF
--- a/src/Whoops/Provider/Silex/WhoopsServiceProvider.php
+++ b/src/Whoops/Provider/Silex/WhoopsServiceProvider.php
@@ -8,7 +8,8 @@ namespace Whoops\Provider\Silex;
 
 use RuntimeException;
 use Silex\Application;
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thank you for your contributions. We definitely do not want to break backward compatibility and remove Silex 1 handler.

As a matter of fact, for a while now we prefer [external modules to integrate Whoops](https://github.com/filp/whoops#installing), Silex 1 is only there for backward compatibility.
It so happens that there already is [an integration with Silex 2](https://github.com/texthtml/whoops-silex) on the list. Please consider forwarding your efforts there!

Thanks again for your good effort.

PR #312, opened 2015-11-01
Wrap PHP7 error exceptions in regular exceptions for compatibility with PHP5

denis-sokolov:
Alternative, simpler, but clunkier implementation of compatibility with PHP 7, #284.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -16,7 +16,7 @@ use Whoops\Handler\HandlerInterface;
 
 class Run
 {
-    const EXCEPTION_HANDLER = "handleException";
+    const EXCEPTION_HANDLER = "handleExceptionFromPHP";
     const ERROR_HANDLER     = "handleError";
     const SHUTDOWN_HANDLER  = "handleShutdown";
 
@@ -293,6 +293,24 @@ class Run
     }
 
     /**
+     * This is a private implementation detail, you are not supposed to call this.
+     */
+    final public function handleExceptionFromPHP($throwable)
+    {
+        if (self::isPHP7Throwable($throwable)) {
+            // Compatibility with Whoops children that expect the exception
+            // to always be a PHP 5 Exception class
+            $throwable = new Exception(
+                get_class($throwable) . ': ' . $throwable->getMessage(),
+                $throwable->getCode(),
+                $throwable
+            );
+        }
+
+        return $this->handleException($throwable);
+    }
+
+    /**
      * Converts generic PHP errors to \ErrorException
      * instances, before passing them off to be handled.
      *
@@ -393,6 +411,14 @@ class Run
         echo $output;
 
         return $this;
+    }
+
+    private static function isPHP7Throwable($something)
+    {
+        return !$something instanceof \Exception &&
+            interface_exists('Throwable', false) &&
+            !is_subclass_of('Throwable', 'Exception') &&
+            $something instanceof \Throwable;
     }
 
     private static function isLevelFatal($level)
--- a/tests/Whoops/ExtendedRunPrePHP7/ExtendedRunPrePHP7Test.php
+++ b/tests/Whoops/ExtendedRunPrePHP7/ExtendedRunPrePHP7Test.php
@@ -1,0 +1,76 @@
+<?php
+namespace Whoops\ExtendedRunPrePHP7;
+
+class Run extends \Whoops\Run {
+    // Simulate users having extended the Run
+    // If we change the typehint in main Run, this will throw
+    public function handleException(\Exception $exception)
+    {
+        return parent::handleException($exception);
+    }
+}
+
+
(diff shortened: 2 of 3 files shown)


hultberg:
:+1:

src/Whoops/Run.php
```diff
@@ -395,6 +410,13 @@ private function writeToOutputNow($output)
         return $this;
     }
 
+    private static function isInternalErrorException($exception)
+    {
+        return class_exists('Error') &&
+            !is_subclass_of('Error', 'Exception') &&
```
staabm: Shouldnt it just check for php7 and use the Error class then (which will always be available in php7 runtime)

staabm: I guess you are trying to check whether the defined Error class was actually provided by php-src instead of a lib/framework, right?

denis-sokolov: Yeah, exactly.
I suppose a PHP version check should be good enough here, but I got used to the idea of feature-checking instead of version-checking from other environments.
In particular, I have no idea how HHVM reports it's version number and which features it implements.

src/Whoops/Run.php
```diff
@@ -292,6 +292,21 @@ public function handleException(Exception $exception)
         return $output;
     }
 
+    final public function handleExceptionFromPHP($exception)
```
staabm: I dont like having this to be part of the public api just to be hookable :-/

denis-sokolov: There, commented it.

staabm:
:+1:

hultberg:
Again, after seeing the latest commits per now :+1:

denis-sokolov:
If anyone has hhvm set up, please investigate the failing test on that environment.

src/Whoops/Run.php
```diff
@@ -395,6 +413,13 @@ private function writeToOutputNow($output)
         return $this;
     }
 
+    private static function isInternalErrorException($exception)
+    {
+        return class_exists('Error') &&
```
staabm: This should not trigger autoload, right?

denis-sokolov: It does. There is even an option `autoload` to this function.

denis-sokolov:
@GrahamCampbell, you're right! Fixing.

src/Whoops/Run.php
staabm: It does but we dont need that

src/Whoops/Run.php
```diff
@@ -293,6 +293,24 @@ public function handleException(Exception $exception)
     }
 
     /**
+     * This is a private implementation detail, you are not supposed to call this.
+     */
+    final public function handleExceptionFromPHP($exception)
```
staabm: we could also do something like

```
// php 5.3 hack
$that= $this;
set_exception_handler(function() {
  $that->{self::EXCEPTION_HANDLER}();
} use ($that));
```

instead of what we currently have
```
set_exception_handler(array($this, self::EXCEPTION_HANDLER));
```

so we dont need this `handleExceptionFromPHP` method to be public

denis-sokolov: I thought about that, but it is so clunky. And for what — to avoid people shooting themselves in the face ignoring the name of the function, its absence from the documentation and the comment forbidding its use? Nah.

staabm: hehe, its clunky absolutely. I am fine with both solutions.

src/Whoops/Run.php
```diff
@@ -395,6 +413,13 @@ private function writeToOutputNow($output)
         return $this;
     }
 
+    private static function isPHP7Throwable($exception)
+    {
+        return class_exists('Throwable', false) &&
```
staabm: `Throwable` is a interface, therefore this should be 

``` php
interface_exists('Throwable', false)
```

instead, see https://3v4l.org/F2JGi

this might be the reason why it fails on travis right now.

staabm: regarding hhvm: this could be tested on 3v4l.org

staabm:
after having a look at the travis hhvm build log it looks like a travis-ci bug for me.

src/Whoops/Run.php
staabm: I guess we should rename the parameter to `$throwable` because we dont know yet it actuall is a `exception`

src/Whoops/Run.php
staabm: shouldnt be the first check be `!($exception instanceof \Exception)` instead?

denis-sokolov:
Thank you, @staabm, for your comments, found good problems with the code.
Now updated.

staabm:
:+1:

denis-sokolov:
With our call to move to a BC break and Whoops 2.0, this is not needed anymore. #284 is now preferred.

PR #314, opened 2015-11-03
Added clipboard.js 

### LINKED ISSUE #313: use clipboard.js instead of zero-clipboard
we should use the new clipboard.js[1] library which uses more modern browser apis to implement copy&paste without flash.

its the successor of the lib we are using right now [2].
Atm whoops loads a SWF file when rendered to be able to provide the copy&paste functionality.

[1] https://github.com/zenorocha/clipboard.js
[2] https://github.com/zeroclipboard/zeroclipboard

only disadvantage I see right now is, that Safari doesnt implement some of the APIs required for the new lib, see https://github.com/zenorocha/clipboard.js#browser-support

prisis:


src/Whoops/Resources/css/whoops.base.css
```diff
@@ -371,61 +370,61 @@ pre.prettyprint, code.prettyprint {
...
-	position: absolute;
-	right: 30px;
-	top: 90px;
+  position: absolute;
+  right: 30px;
+  top: 90px;
 }
 
 #help-framestack {
-	position: absolute;
```
staabm: you changed a lot of whitespace, which creates unnecessary noise in this PR

staabm:
you used 1.5.2 but 1.5.3 is the lastest release of clipboard.js

prisis:
Hey @staabm im using the lastest [clipboard.js](https://github.com/zenorocha/clipboard.js) ;) 1.5.3 is only a package.json change

src/Whoops/Resources/css/whoops.base.css
prisis: Yes i did, because they were all mixed...

staabm: IMO this should be done in either a separate commit or even PR

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -469,3 +468,59 @@ kbd {
     width: 32%;
   }
 }
+
+/* Tooltipp */
```
staabm: useless comment

denis-sokolov:

As @staabm pointed out, having a lot of whitespace changes in the same pull request creates a lot of problems. It causes unexpected conflicts, it does not allow us to consider and discuss the changes separately.

Please consider rebasing the PR to only include changes related to clipboard.js.
Whitespace cleanup can be done separately, later, after ensuring it does not break anyone's workflow.

Thank you!

prisis:
i make a new pull request.

zenorocha:
lol

PR #315, opened 2015-11-04
load prettify asynchronously to speedup rendering

staabm:
Inspired by state of the art frontend performance best practices.

### DIFF
--- a/src/Whoops/Resources/js/whoops.base.js
+++ b/src/Whoops/Resources/js/whoops.base.js
@@ -1,5 +1,20 @@
 Zepto(function($) {
-  prettyPrint();
+  // a jQuery.getScript() equivalent to asyncronously load javascript files
+  // credits to http://stackoverflow.com/a/8812950/1597388
+  var getScript = function(src, func) {
+    var script = document.createElement('script');
+    script.async = 'async';
+    script.src = src;
+    if (func) {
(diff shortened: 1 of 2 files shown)


staabm:
This might sometimes lead to code beeing highlighted a bit later, but will lead to earlier execution of the Zepto related JS code in `whoops.base.js`, so user interaction with the page is possible earlier.

denis-sokolov:
Could you put it into a variable instead of sticking it on Zepto?

```js
var getScript = function(){
};
```

And fix the inconsistent quote characters.

A great addition, by the way!

staabm:
@denis-sokolov good point, adjusted per review.

denis-sokolov:
There are more quote characters in that code!

staabm:
@denis-sokolov oh sorry, missed it.

should I fix the quotes in the newly added code or all over the files?

denis-sokolov:
Only the newly added.
It could be a reasonable idea to clean up the entire codebase, but definitely as a separate PR.
In addition, because that involves touching most files, it’s best to be done when there are no active big side branches to avoid conflicts.
At the moment we have that restyling branch, so I would postpone any such cleanup work.

staabm:
@denis-sokolov alright. fixed the quotes.

PR #317, opened 2015-11-08
Replaced ZeroClipboard with clipboard | fixes #313

### LINKED ISSUE #313: use clipboard.js instead of zero-clipboard

prisis:


denis-sokolov:
Seems good to me. @staabm, do you see any problems?

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -469,3 +468,58 @@ kbd {
...
+  right: 50%;
+  margin-top: 5px
+}
+.tooltipped-s:before {
+  top: auto;
+  right: 50%;
+  bottom: -5px;
+  margin-right: -5px;
+  border-bottom-color: rgba(0, 0, 0, 0.8)
+}
```
GrahamCampbell: missing newline

prisis: done.

staabm:
Looks good, didnt test it though

PR #318, opened 2015-11-17
Css clean up.

prisis:
@filp i hope, is okay like this.

#292


filp:
To be honest I'm not a big fan of this - 2 spaces for CSS seems to be the standard pretty much everywhere I've seen, and removing the nesting hurts readability in this case.

denis-sokolov:
I do agree with filp, there is no clear value added in these changes, merely changing one preference for another. We will not be merging this.

That said, it is great that you have suggested this and an opportunity to something like this is always beneficial. Thank you for helping, @prisis.

PR #320, opened 2015-11-21
Optionally use symfony/var-dumper to pretty print var-dumping

staabm:
Closes #308 

Doesnt look that pretty right now:
vardumper

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -458,3 +458,8 @@ kbd {
   margin-right: -5px;
   border-bottom-color: rgba(0, 0, 0, 0.8)
 }
+
+pre.sf-dump {
+    padding: 0px !important;
+    margin: 0px !important;
+}
--- a/src/Whoops/Util/TemplateHelper.php
+++ b/src/Whoops/Util/TemplateHelper.php
@@ -6,6 +6,10 @@
 
 namespace Whoops\Util;
 
+use Symfony\Component\VarDumper\Cloner\VarCloner;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
+use Symfony\Component\VarDumper\Dumper\HtmlDumper;
+
 /**
  * Exposes useful tools for working with/in templates
  */
@@ -56,6 +60,44 @@ class TemplateHelper
             "@([A-z]+?://([-\w\.]+[-\w])+(:\d+)?(/([\w/_\.#-]*(\?\S+)?[^\.\s])?)?)@",
             "<a href=\"$1\" target=\"_blank\">$1</a>", $escaped
         );
+    }
+
+    /**
+     * Format the given value into a human readable string.
+     *
+     * @param  mixed $value
+     * @return string
+     */
+    public function dump($value)
+    {
+        if (class_exists('Symfony\Component\VarDumper\Cloner\VarCloner')) {
+            static $dumper = null;
+
+            // re-use the same var-dumper instance, so it won't re-render the global styles/scripts on each dump.
+            if (!$dumper) {
+                $dumper = new HtmlDumper();
+
+                $styles = array(
+                    'default' => '',
+                    'num' => '',
+                    'const' => '',
+                    'str' => '',
+                    'note' => '',
+                    'ref' => '',
+                    'public' => '',
+                    'protected' => '',
+                    'private' => '',
+                    'meta' => '',
+                    'key' => '',
+                    'index' => '',
+                );
+                $dumper->setStyles($styles);
+            }
+
+            $cloner = new VarCloner();
+            return $dumper->dump($cloner->cloneVar($value));
+        }
+        return print_r($value, true);
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,9 @@
     "require-dev": {
         "mockery/mockery": "0.9.*"
     },
+    "suggest": {
+        "symfony/var-dumper": "^2.6"
+    },
     "autoload": {
         "psr-4": {
             "Whoops\\": "src/Whoops/"
--- a/src/Whoops/Resources/views/env_details.html.php
+++ b/src/Whoops/Resources/views/env_details.html.php
@@ -17,7 +17,7 @@
             <?php foreach ($data as $k => $value): ?>
               <tr>
                 <td><?php echo $tpl->escape($k) ?></td>
-                <td><?php echo $tpl->escape(print_r($value, true)) ?></td>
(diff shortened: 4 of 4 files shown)


src/Whoops/Util/TemplateHelper.php
```diff
@@ -57,6 +57,19 @@ public function escapeButPreserveUris($raw)
...
     }
+	
+    /**
+     * Format the given value into a human readable string
+     *
+     * @param  mixed $value
+     * @return string
+     */
+    public function dump($value) {
+        if (class_exists('\Symfony\Component\VarDumper\VarDumper')) {
```
GrahamCampbell: don't need a leading slash here

GrahamCampbell: infact, it's technically wrong to include it, but it will still work because php trims it under the hood

src/Whoops/Util/TemplateHelper.php
```diff
@@ -57,6 +57,19 @@ public function escapeButPreserveUris($raw)
...
+            return \Symfony\Component\VarDumper\VarDumper::dump($value);
```
GrahamCampbell: maybe better to import this - importing doesn't fail if the class doesn't exist - infact php doesn't even attempt to autoload it at that point

src/Whoops/Util/TemplateHelper.php
```diff
@@ -57,6 +57,19 @@ public function escapeButPreserveUris($raw)
...
         );
     }
+	
+    /**
+     * Format the given value into a human readable string
+     *
+     * @param  mixed $value
+     * @return string
+     */
+    public function dump($value) {
```
GrahamCampbell: cs

src/Whoops/Util/TemplateHelper.php
staabm: so whats the benefit of importing it? I would need `class_exists` here, no matter it is imported or not, right?

src/Whoops/Util/TemplateHelper.php
staabm: fixed

src/Whoops/Util/TemplateHelper.php
staabm: fixed

src/Whoops/Util/TemplateHelper.php
GrahamCampbell: importing is just a cs thing - doesn't affect anything else ;)

src/Whoops/Util/TemplateHelper.php
```diff
@@ -57,6 +57,20 @@ public function escapeButPreserveUris($raw)
...
+    /**
+     * Format the given value into a human readable string
+     *
+     * @param  mixed $value
+     * @return string
+     */
+    public function dump($value)
+    {
+        if (class_exists('Symfony\Component\VarDumper\VarDumper')) {
+            return Symfony\Component\VarDumper\VarDumper::dump($value);
```
GrahamCampbell: missing slash, or import it

denis-sokolov:
Hm. Yes, not that awesome. What if we only use it for non scalar values?

staabm:
I guess we need to customize the css provided by the dumper.. Will see after the redesign landed to get a full picture

GrahamCampbell:
> What if we only use it for non scalar values?

I think symfony copes ok with that. Laravel uses it in it's dd command.

staabm:
@GrahamCampbell I guess denis just proposed it to have less ugly places :-)

src/Whoops/Util/TemplateHelper.php
```diff
@@ -59,6 +59,20 @@ public function escapeButPreserveUris($raw)
     }
 
     /**
+     * Format the given value into a human readable string
```
GrahamCampbell: missing fullstop

denis-sokolov:
@staabm, exactly.

staabm:
Added a few lines of CSS and override some of the `HtmlDumpers` styles to get something which integrates better with our theme:

vardumper

So ATM this PR adds all the expand/collapse and casters functionality, without any visual stuff.
As no visuals are defined, it should also work with the redesig.

Opened a upstream issue on https://github.com/symfony/symfony/issues/16618 which targets to ease css customization of the `HtmlDumper`.

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -426,3 +426,8 @@ kbd {
     padding: 3px 5px;
     vertical-align: middle;
 }
+
+pre.sf-dump {
+    padding: 0px !important;
```
staabm: its ugly, but atm the only way to override the styles the symfony/var-dumper renders into the body using `<style>`-tags

src/Whoops/Util/TemplateHelper.php
```diff
@@ -59,6 +63,42 @@ public function escapeButPreserveUris($raw)
...
+     * Format the given value into a human readable string.
+     *
+     * @param  mixed $value
+     * @return string
+     */
+    public function dump($value)
+    {
+        if (class_exists('Symfony\Component\VarDumper\Cloner\VarCloner')) {
+            $cloner = new VarCloner();
+            if ('cli' === PHP_SAPI) {
```
GrahamCampbell: not very reliable

GrahamCampbell: maybe better for whatever's calling the dump function to pass it's context through

denis-sokolov: I think it's plenty reliable. The entire internet of PHP devs use this, it won't be broken.
At least for our purposes it's not vital enough to worry about the risk.

denis-sokolov:
@staabm, let me know when you've updated this to the state where you want to merge.

src/Whoops/Util/TemplateHelper.php
GrahamCampbell: > The entire internet of PHP devs use this

Not true, really. I'd never use it in that context. I'd always have that part at a higher level of abstaction.

staabm: To be honest, I copied this part out of `VarDumper::dump` without further thinking.. Atm we only use this viewHelper in web context, so actually we can drop the cli-sapi case.

GrahamCampbell: > so actually we can drop the cli-sapi case.

:+1:

staabm:
@denis-sokolov This is ready to merge. Not the most elegant solution, but introducing a new subclass of HtmlDumper within Whoops just for a bit more customization seems like a bit overkill to me.

This works as intendend. Also tested on the new re-designed PrettyPageHandler.

PR #321, opened 2015-11-21
remove onlyForAjaxRequests in JsonResponseHandler and move its logic into a re-usable Whoops\isAjaxRequest helper function

### LINKED ISSUE #319: Add onlyForAjaxRequests to PlainTextHandler
The `JsonResponseHandler` offers the `onlyForAjaxRequests` configuration option so that you can get the pretty HTML page when directly navigating to a URL in your browser and a JSON response (which I guess you can parse and display in some sensible way if you're so inclined) in response to XHR requests.

However, if you *don't* want to write any code to parse the JSON response and just want something pretty and readable in the 'Network' tab of your dev tools, the `JsonResponseHandler` sucks; especially if you include the traceback, it becomes almost unreadable. The `PlainTextHandler` would be vastly superior for this use case.

The pretty HTML error responses also aren't ideal if you want to read them in the network tab, since they don't look great in a small window and none of the interactivity works. 

So why not add `onlyForAjaxRequests` to the `PlainTextHandler` and have it behave just like it does on the `JsonResponseHandler`? This would make it usable for people like me who just want a pretty plaintext response to their AJAX calls that they can read in the devtools, but still want to be able to get the cool HTML error pages when navigating to a URL directly.

staabm:
closes #319.

As requested I only implemented it in PlainTextHander. Not sure whether we should add it in all handlers.

As it does not make sense in non-http handlers, I would suggest to move the Logic into a `trait` ... thoughts on that?

### DIFF
--- a/src/Whoops/Handler/JsonResponseHandler.php
+++ b/src/Whoops/Handler/JsonResponseHandler.php
@@ -21,11 +21,6 @@ class JsonResponseHandler extends Handler
     private $returnFrames = false;
 
     /**
-     * @var bool
-     */
-    private $onlyForAjaxRequests = false;
-
-    /**
      * @param  bool|null  $returnFrames
      * @return bool|$this
      */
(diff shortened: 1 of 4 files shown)


denis-sokolov:
Perhaps instead we remove it from handlers at all and expose a helper?
The client code will read

```php
if (Whoops\isAjaxRequest()){
  $run->addHandler($myHandler)
}
```

Not a particular degradation compared to

```php
$handler->onlyForAjax();
$run->addHandler($handler);
```

The code design improves much, though.

staabm:
Since we get a BC break on master anyway we could change the API as suggested...

src/Whoops/Handler/JsonResponseHandler.php
```diff
@@ -21,11 +21,6 @@ class JsonResponseHandler extends Handler
     private $returnFrames = false;
 
     /**
-     * @var bool
```
staabm: @denis-sokolov I assume a major version bump, therefore broke BC to cleanup the handler. In case you wanne preserve BC give me a ping, so I will re-add and deprecate this stuff instead.

denis-sokolov:
Would you please also document this in the user-facing documentation?

ExplodingCabbage:
> I would suggest to move the Logic into a `trait` ... thoughts on that?

This would be the cleanest thing to do, but unfortunately traits were only added in PHP 5.4, so you'd be bumping Whoops's minimum PHP version to 5.4 at a time when 5.3 is [the most popular version of PHP](http://w3techs.com/technologies/details/pl-php/5/all) deployed on the web. Alas, this is probably a bad idea.

staabm:
@denis-sokolov docs added.

ExplodingCabbage:
If y'all go with this approach, one thing worth considering is whether you also want to move the somewhat analogous `onlyForCommandLine()` method of PlainTextHandler out to a standalone function.

staabm:
> If y'all go with this approach, one thing worth considering is whether you also want to move the somewhat analogous onlyForCommandLine() method of PlainTextHandler out to a standalone function.

very good point, didnt realized we already have another function doing similar stuff. In case denis agrees on the approach we should move this method in a separate PR.

PR #325, opened 2015-11-22
Only support maintained versions of PHP

GrahamCampbell:
What do you make of this? This matches what Laravel and Symfony has done. PHPUnit has gone one step further, and actually only supports PHP 5.6+ now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,7 @@
 language: php
 
 php:
-  # 5.3.0 is not supported by Travis
-  - 5.3.3
-  - 5.3
-  - 5.4
+  - 5.5.9
   - 5.5
   - 5.6
   - 7.0
@@ -12,10 +9,7 @@ php:
 
 sudo: false
 
-before_script:
-  - composer self-update
-  # --prefer-source is required to avoid hitting GitHub API limit:
-  # https://github.com/composer/composer/issues/1314
+install:
   - composer install --no-interaction --prefer-source
   - if [ $TRAVIS_PHP_VERSION = "5.6" ]; then PHPUNIT_FLAGS="--coverage-clover build/logs/clover.xml"; fi
 
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0"
+        "php": ">=5.5.9"
     },
     "require-dev": {
         "mockery/mockery": "0.9.*"


hultberg:
This means that developers with Debian Wheezy can not use Whoops because it has PHP 5.4.45. On some server hosts they have still not updated to jessie or a higher ubuntu version. Ubuntu 14.04 LTS has PHP 5.5.9. I also checked fedora and arch linux but no worries there, they all have PHP 5.6+. When it comes to CentOS, they do not have a online package browser so I'm unable to check.

I do support removing support for older PHP versions so those who need to use a lower PHP version just have to use a different version of Whoops.

Refs:
* https://packages.debian.org/wheezy/php5
* http://packages.ubuntu.com/trusty/php5

.travis.yml
```diff
@@ -12,10 +9,7 @@ php:
 
 sudo: false
 
-before_script:
```
hultberg: Why removing self-update of composer? Just wondering.. :-P

GrahamCampbell: Because there's no point. It just wastes time.

GrahamCampbell:
People using OSs with older versions of PHP should either upgrade the PHP version, or use an older version of whoops.

Alternatively, if we think going to 5.5 is too soon, we could always just bump to 5.4.

.travis.yml
```diff
@@ -1,10 +1,7 @@
 language: php
 
 php:
-  # 5.3.0 is not supported by Travis
-  - 5.3.3
-  - 5.3
-  - 5.4
+  - 5.5.9
```
staabm: Why 5.5.9?

staabm:
I would opt for 5.4+

.travis.yml
GrahamCampbell: That's the version that comes with ubuntu 14.04. It's also kind of pointless to allow 5.5.0 since that's older than the lots of 5.4.x releases.

GrahamCampbell: Also, Laravel and Symfony both did that.

hultberg:
@GrahamCampbell A good point. After researching I did not get as worried as I initially was because major OS that provide PHP is at minimum 5.4 as with wheezy. But, if a developer is still using wheezy I would just tell the person too stop lagging behind, srz.

.travis.yml
GrahamCampbell: Why test on 5.5.9? Because that's now the lowest version we support in the composer.json, so good to assert that it actually works.

denis-sokolov:
I don't see any specific value in cutting off people on purpose without any practical benefit.
If we come up with a new feature that relies on new PHP and breaks with older versions of PHP, sure, let's drop the support.
But for no reason? Let's rather not.

GrahamCampbell:
How about bumping to 5.3.3? At least we can actually test if that works.

denis-sokolov:
@GrahamCampbell, absolutely!

PR #330, opened 2015-11-22
Forbid extending Run

denis-sokolov:


staabm:
Any insights why we would want that?

denis-sokolov:
I do not see a use case where this is beneficial, and it does restrict us a lot in changing the implementation. That is, I want to minimize API surface area, encapsulate.
Of course, if you let me know why would someone need to extend the class, I'm all ears.

staabm:
Should we also change protected things to private in this step?

denis-sokolov:
Optionally, yes.
Protected in final is equivalent to private anyway and I wanted to keep the PR small to keep the discussion going.

staabm:
Yeah it makes sense in case we later on need to drop the `final` to have the least possible api surface.

mahagr:
In my use case Whoops is running in multiple different platforms including Joomla, WordPress and Grav and because of those CMSes have their own error handlers, I have made adjustments to turn whoops off for any code outside of my application scope (opposite to the Whoops filters). This is because of both Joomla and WP has so much of really badly written code that Whoops would break half of the sites out there.

As this behaviour isn't really needed for most of the applications that use Whoops, I've just extended the class to add this new behavior. I can of course make a pull request of my changes, but I'm not sure if they fit to the project, which is why I extended the class myself.

Here is the code in case if you are interested to see it; I can make pull request if needed:

https://github.com/gantry/gantry5/blob/develop/src/classes/Gantry/Component/Whoops/Run.php

Without the ability to extend the `Run` class it would be impossible to make adjustments like this to cope the underlying platform. I agree that changes in Whoops will break my code, but I have always made checks to ensure that my class stays updated and will match the version of Whoops I have updated via composer.

denis-sokolov:
I'm not sure I understand. Most of the file you linked to is made in Whoops. (Displaying a bad amount of copy and paste). What's the custom code in that class?

mahagr:
The class is basically doing the opposite to your code. Instead of saying which files to ignore, it says which files cause Whoops to run. For other files it falls back to the previously assigned error handler.

The code is almost the same as I just added my logic into existing methods.

staabm:
coudn't you have implemented this with as a composite in a cleaner way (instead via inheritance)?

denis-sokolov:
Yeah, there is no need to have an inheritance. A simple condition would suffice:

```php
set_error_handler(function(){
  if (filetohandle()) {
    $whoops->handle()
  }
});
```

Better yet, indeed, we can add whitelisting into Whoops core!

mahagr:
Well yes, I didn't even think of registering a new error handler to do the whitelisting as my original approach provided me a cleaner API to do the same. :)

I can prepare a pull request on allowing user to use whitelisting instead of blacklisting if you want me to add that feature. It also adds a call to the previous error handler to make it more compatible to external code that may have set its own error handling function.

I'm just about to leave to my long waited vacation, so I'll probably get back to you early next year.

denis-sokolov:
If you add a feature cleanly, in a very symmetrical way to how the blacklisting works now, then great!

mahagr:
I will take the challence; will get back to you after my vacation. :)

PR #332, opened 2015-11-22
Let onlyForAjaxRequests be called on any handler

ExplodingCabbage:
Perhaps this is a more useful approach than https://github.com/filp/whoops/pull/321 ?

This pull request has two commits; it's possible you'll want to merge the first and reject the second.

The first commit moves `onlyForAjaxRequests()` to the base `Handler()` class and has `Run()` call it on all Handlers. The advantage of this over https://github.com/filp/whoops/pull/321 is that it doesn't break backwards compatibility for people who are already calling `onlyForAjaxRequests()` and it keeps all Handler configuration done by calling Handler methods instead of needing to pull some conditional logic out of the handler. Disadvantages I can see are that it breaks backwards compatibility for anyone who was implementing `HandlerInterface` themself without inheriting from `Handler` (which is probably a bad idea in the first place) and perhaps makes it less obvious how to add only-for-Ajax behaviour to a closure (since that will require explicitly creating a `CallbackHandler` rather than relying on the automatic creation caused by passing a closure to `pushHandler()`.

The second commit adds a teeny hack to `PlainTextHandler` to make it play nicer with the use case I described in https://github.com/filp/whoops/issues/319. Basically, `PlainTextHandler` only outputs to the command line by default, but if we're calling `onlyForAjaxRequests(true)` on it then plainly that's not the behaviour we want, so I made the `onlyForAjaxRequests(true)` call automatically turn off the command-line-only switch. The upside of this is that it removes a confusing trap that people will stumble into when using the PlainTextHandler for Ajax calls. (At least, I was confused when testing!) The downside is that if somebody has got their PlainTextHandler working with Ajax calls by unknowingly relying on this magic behaviour and then they decide to make the PlainTextHandler their sole handler for using everything by deleting their other handlers and removing the `->onlyForAjaxRequests(true)` line from their code, then suddenly their handler won't work on the web at all, and they'll have an *even more confusing* debugging task.

I don't see an easy way to dodge these tradeoffs and I think the calls I've made in both commits are probably the best ones on balance, but it's your call on both.

I don't have enough PHPUnit or Mockery knowledge to feel confident adding tests so I just did the minimum test-modification necessary to keep the current tests passing. I also manually tested that `->onlyForAjaxRequests()` works properly on the PlainTextHandler now.

### DIFF
--- a/docs/API Documentation.md
+++ b/docs/API Documentation.md
@@ -117,6 +117,11 @@ HandlerInterface::setInspector(Whoops\Exception\Inspector $inspector)
 // Sets the Exception for this handler to handle
 HandlerInterface::setException(Exception $exception)
  #=> null
+ 
+// Should output only be sent if the current request is an
+// AJAX request?
+HandlerInterface::onlyForAjaxRequests(bool $yes = null)
+ #=> bool
 ```
 
 ## <a name="inspector"></a> `Whoops\Exception\Inspector`
@@ -305,11 +310,6 @@ The `JSON` body has the following format:
 // Should detailed stack trace output also be added to the
 // JSON payload body?
 JsonResponseHandler::addTraceToOutput(bool $yes = null)
- #=> bool
-
-// Should output only be sent if the current request is an
-// AJAX request?
-JsonResponseHandler::onlyForAjaxRequests(bool $yes = null)
  #=> bool
 
 JsonResponseHandler::handle()
--- a/src/Whoops/Handler/Handler.php
+++ b/src/Whoops/Handler/Handler.php
@@ -38,6 +38,11 @@ abstract class Handler implements HandlerInterface
      * @var Exception $exception
      */
     private $exception;
+    
+    /**
+     * @var bool
+     */
+    private $onlyForAjaxRequests = false;
 
     /**
      * @param Run $run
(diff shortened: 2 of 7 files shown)


denis-sokolov:
First of all, thank you for your work.

Since #319, in #321 we have moved in a different direction. Now you submit a code proposal that goes back to the scenario discussed in #319, directly competing with code in #321. One way or another I will need to reject some code. :(

Now, I see advantages of both approaches, and I need to ponder a bit.

ExplodingCabbage:
Yep, I just wanted to chuck you a candidate for this approach so that the choice would be open to you. No hard feelings if you reject this one!

staabm:
design wise #321 is the cleaner approach as it keeps our handlers free from this SAPI specific stuff and let the caller decide when to use what.

Main disadvantage of this approach (which is the same we had in #321 before we changed to what we have now) that all handlers in the future have  a `onlyForAjaxRequests`, even for handlers where it might make not sense at all. Second disadvantage is, that in case we get some additional method, like e.g. `onlyForWebsocketRequests` this will further pollute the whole api.

denis-sokolov:
After all, I'd rather simplify our system. The user code is arguably simpler as well:

```php
$handler = new JsonResponseHandler();
$handler->onlyForAjaxRequests();
$run->addHandler($handler);
```

```php
if (isAjaxRequest())
  $run->addHandler(new JsonResponseHandler());
```

PR #333, opened 2015-11-23
[WIP] Refactor method onlyForAjaxRequests in JsonResponseHandler

hultberg:
This method has two jobs, and this PR will split that into one and add a new method to set the flag for onlyForAjaxRequest with `setOnlyForAjaxRequest`. It is a good practice to have one method do one thing, not multiple. :smile: 

I spotted this in PR #332 when looking at the changes there.

If PR #332 is accepted I would need to update my PR.

Cheers

### DIFF
--- a/src/Whoops/Handler/JsonResponseHandler.php
+++ b/src/Whoops/Handler/JsonResponseHandler.php
@@ -40,15 +40,22 @@ class JsonResponseHandler extends Handler
     }
 
     /**
-     * @param  bool|null $onlyForAjaxRequests
+     * Get if handler is only for ajax requests.
+     *
      * @return null|bool
      */
-    public function onlyForAjaxRequests($onlyForAjaxRequests = null)
+    public function onlyForAjaxRequests()
     {
-        if (func_num_args() == 0) {
-            return $this->onlyForAjaxRequests;
-        }
+        return $this->onlyForAjaxRequests;
+    }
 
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thank you for your contribution, @hultberg. I definitely see benefit in making the code flows clearer.
There are three reasons I will not merge this in.

Firstly, consistency. [All our methods](https://github.com/filp/whoops/blob/master/src/Whoops/Run.php) work this way, see `allowQuit`, `sendHttpCode`, `writeToOutput`.

Secondly, backwards compatibility. These are big changes to the API, they better be worth it.

Thirdly, they are not worth it. The current code is not doing two unrelated, confusing and intermingled logic scenarios at once. This is simple function overloading, separated by a clear conditional in the implementation. Using or not using such approach is really 80% style and only 20% any objective advantage or disadvantage.

Again, thank you for contributing regardless, and this is a clean and smart PR.

hultberg:
As we where bumping to a new major version I wanted to introduce some BC changes. :grin:

denis-sokolov:
A major version does allow us to make BC changes, but they still have a cost associated with them.
The more BC changes, the harder the transition for users.

Our main breaking change with extending Run signature with PHP 7 Errors is probably meaningless for majority of users, which means they will be able to upgrade without any modifications to their code at all. We should try to keep that.

PR #334, opened 2015-11-23
Move SOAP handler out

denis-sokolov:


staabm:
We should declare those special handlers in composer.json as suggested package

denis-sokolov:
You’re right.

PR #337, opened 2015-11-29
Added support for printing frame args in the framestack.

### LINKED ISSUE #15: Display arguments of failed function calls
Arguments used for each function is stocked in the backtrace, it would be nice if those were displayed somewhere.

staabm:
This adds printing of frame args, in case symfony/var-dumper is present.
When the sf component is missing the page looks as before. I took this path, because our default `print_r` var-dumping produces a big mess in this small panel on the left.

Looks like
http://i.imgur.com/PzebmFI.png
http://i.imgur.com/Ortpdmi.png

Should fix #15.

In case https://github.com/symfony/symfony/issues/16741 gets fixed, we might also get away from the default-expanded objects.

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -133,6 +133,15 @@ header {
       font-size: 14px;
     }
 
+    .frame-arg {
+      margin-left: 20px;
+      display: block;
+    }
+
+    .frame-arg-separated .sf-dump::after {
+      content: ',';
+    }
+
     .frame-index {
       font-size: 11px;
       color: #a29d9d;
--- a/src/Whoops/Resources/views/env_details.html.php
+++ b/src/Whoops/Resources/views/env_details.html.php
@@ -17,7 +17,7 @@
             <?php foreach ($data as $k => $value): ?>
               <tr>
                 <td><?php echo $tpl->escape($k) ?></td>
-                <td><?php echo $tpl->escape($tpl->dump($value)) ?></td>
+                <td><?php echo $tpl->dump($value) ?></td>
               </tr>
             <?php endforeach ?>
             </table>
--- a/src/Whoops/Resources/views/frame_list.html.php
+++ b/src/Whoops/Resources/views/frame_list.html.php
@@ -7,6 +7,7 @@
         <span class="frame-index"><?php echo (count($frames) - $i - 1) ?></span>
         <span class="frame-class"><?php echo $tpl->escape($frame->getClass() ?: '') ?></span>
         <span class="frame-function"><?php echo $tpl->escape($frame->getFunction() ?: '') ?></span>
+        <?php echo $tpl->dumpArgs($frame); ?>
       </div>
 
     <span class="frame-file">
@@ -14,4 +15,4 @@
    --><span class="frame-line"><?php echo (int) $frame->getLine() ?></span>
     </span>
   </div>
-<?php endforeach ?>
+<?php endforeach; ?>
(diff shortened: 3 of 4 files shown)


src/Whoops/Resources/css/whoops.base.css
```diff
@@ -133,6 +133,15 @@ header {
       font-size: 14px;
     }
 
+    .frame-arg {
+      margin-left: 20px;
+      display: block;
+    }
+
+    .frame-arg-separated .sf-dump::after {
```
staabm: I dont know why, but atm the arg separating `,` is rendered in a separate line. could get the css to render it after the actual arg.

staabm:
visually not the nicest thing but maybe @filp or someone else has an idea how it could more beautifull

src/Whoops/Resources/views/frame_list.html.php
```diff
@@ -7,6 +7,17 @@
         <span class="frame-index"><?php echo (count($frames) - $i - 1) ?></span>
         <span class="frame-class"><?php echo $tpl->escape($frame->getClass() ?: '') ?></span>
         <span class="frame-function"><?php echo $tpl->escape($frame->getFunction() ?: '') ?></span>
+
+        <?php if (class_exists('Symfony\Component\VarDumper\Cloner\VarCloner')):
```
denis-sokolov: This conditional is too specific. Perhaps we could hide this logic inside the template. `if ($tpl->fancyDumperAvailable())`?

staabm: Yeah, guessed. Will adjust after we approve all other details

denis-sokolov:
Great job on moving this task further! I approve.
Let me know whether or not you plan to tweak the looks and the conditional further.

src/Whoops/Resources/views/frame_list.html.php
staabm: We could also just add `$tpl->dumpFrameArgs()` which returns an empty string when no dumper is available

staabm:
@denis-sokolov you are welcome to take over the visuals in case you like

staabm:
@denis-sokolov just pushed a refactored patch. As of now, the html for the frameargs is rendered in the `TemplateHelper`

src/Whoops/Util/TemplateHelper.php
```diff
@@ -70,37 +99,56 @@ public function escapeButPreserveUris($raw)
...
-                );
-                $dumper->setStyles($styles);
-            }
+        $dumper = $this->getDumper();
 
+        if ($dumper) {
             $cloner = new VarCloner();
-            return $dumper->dump($cloner->cloneVar($value));
+            $output = '';
+            $dumper->dump($cloner->cloneVar($value),  function ($line, $depth) use (&$output) {
```
staabm: this change seems unrelated, but actually it fixes the `TemplateHelper->dump` method to actually return the dumped html instead of echo'ing it.

src/Whoops/Resources/views/env_details.html.php
```diff
@@ -17,7 +17,7 @@
             <?php foreach ($data as $k => $value): ?>
               <tr>
                 <td><?php echo $tpl->escape($k) ?></td>
-                <td><?php echo $tpl->escape($tpl->dump($value)) ?></td>
+                <td><?php echo $tpl->dump($value) ?></td>
```
staabm: as the string is now properly returned, it reveals a bug here... the escaping is not required/correct, as `dump()` returns html.

staabm:
Will continue working on this next week

damnedest:
I think that better show variables dumb above environment because sometimes list of variables/their dumb can be huge.

damnedest:
@staabm buddy, you till working on this?

staabm:
@damnedest any support is welcome. We could collaborate on this in my fork, if you like.

Just propose a PR to my fork.

staabm:
> I think that better show variables dumb above environment because sometimes list of variables/their dumb can be huge.

Feel free to propose whichever design you think would fit best. We could discuss alternatives when you provide screenshots or similar

damnedest:
Great, I'll try to give some suggestions on this, coz this is only one thing I realy need from Exceptions =)

staabm:
Closing in favour of https://github.com/filp/whoops/pull/387

PR #338, opened 2015-11-29
Removed PlainTextHandler->(onlyForCommandLine|outputOnlyIfCommandLine…

### LINKED ISSUE #335: Refactor PlainTextHandler->onlyForCommandLine out into function
consider move the somewhat analogous onlyForCommandLine() method of PlainTextHandler out to a standalone function.

Like https://github.com/filp/whoops/pull/321

staabm:
…). Added Whoops\isCommandLine() as a more flexible successor.

The actual great thing about this change is, that we drop a lot of unecessary api but staying as flexible as before.

fixes #335

### DIFF
--- a/docs/API Documentation.md
+++ b/docs/API Documentation.md
@@ -14,6 +14,7 @@
 
 ### Core Functions:
 - [`Whoops\isAjaxRequest()`](#fn-ajax) - Determines whether the current request was triggered by XMLHttpRequest
+- [`Whoops\isCommandLine()`](#fn-cli) - Determines whether the current request was triggered via php commandline interface (CLI)
 
 
 # Core Classes:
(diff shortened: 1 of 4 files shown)


docs/API Documentation.md
```diff
@@ -408,10 +409,33 @@ PrettyPageHandler::handle()
...
+```php
+// Use a certain handler only in php cli
+if (Whoops\isCommandLine()){
+  $run->addHandler($myHandler)
+}
+```
+
+```php
+/*
+Output the error message only if using command line.
```
staabm: I added this sample, as it 1:1 reflects the behavior of the deleted `outputOnlyIfCommandLine`

tests/Whoops/Handler/PlainTextHandlerTest.php
```diff
@@ -311,7 +242,7 @@ public function testReturnsWithFramesOutput()
                 'Whoops\Handler\PlainTextHandlerTest',
                 'getException',
                 __FILE__,
-                61
+                55
```
staabm: We should not rely on the line a method is defined within a file...

PR #339, opened 2015-11-29
Use rounded corners for frame-index.

staabm:
Personally I think the circles used around the frame-index are not consistent with the rest of the new layout. In all other places we use rectangles with rounded corners.

This PR also introduces small rectangles instead of the circles.

BEFORE
round

AFTER
rounded


staabm:
//cc @filp

denis-sokolov:
I shall question the entire point of numbers altogether. But that is unrelated to the PR. The PR itself gets my approval.

staabm:
@denis-sokolov rwady to land?

PR #343, opened 2015-12-09
Fix gh-pages JsonResponseHandler example

ExplodingCabbage:
This is correct syntax for the master branch; it shouldn't go live until the changes currently on master are released.


staabm:
:+1:

denis-sokolov:
We should instead hurry with releasing new version. :)

PR #345, opened 2015-12-11
Fix handling of Content-Type headers

ExplodingCabbage:
There was previously a seemingly redundant check for the existence of `\Whoops\Utils\Misc` in `PlainTextHandler`; I've purged it, although perhaps there's some logic to it that I'm not seeing?

I've also made the `PrettyPageHandler` spit out `Content-Type: text/html` so that if I cause an error after I've already emitted another `Content-Type` header, the pretty HTML error page can still render.

### DIFF
--- a/src/Whoops/Handler/PlainTextHandler.php
+++ b/src/Whoops/Handler/PlainTextHandler.php
@@ -258,8 +258,7 @@ class PlainTextHandler extends Handler
             return Handler::DONE;
         }
 
-        if (class_exists('\Whoops\Util\Misc')
-            && \Whoops\Util\Misc::canSendHeaders()) {
+        if (\Whoops\Util\Misc::canSendHeaders()) {
             header('Content-Type: text/plain');
         }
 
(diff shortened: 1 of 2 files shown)


hultberg:
I think you should have spilt this PR into two different ones because of the seemingly two different changes.

ExplodingCabbage:
@hultberg I disagree. If https://github.com/ExplodingCabbage/whoops/commit/1467640a51dd1c7c00ef4978919bd2f413462a33 is bad (because there's some reason for that check to be there), then https://github.com/ExplodingCabbage/whoops/commit/d7d89b0cce35b7dff10aae14d5b075c3c6acd610 is probably *also* bad (since the circumstances are basically identical and if the `class_exists` check was needed in `PlainTextHandler` then it's probably also needed in `PrettyPageHandler`). By submitting the two changes together I'm hoping to draw reviewers' attention to this and reduce the chance that the second commit will slip through with some defect that I don't understand.

I guess I could've submitted the first commit as one pull request, waited for it to be accepted, and then come back later to submit the second, but this seemed more straightforward.

denis-sokolov:
Thanks for your contribution, it's great. Your reasoning in your comment is also very solid.
Failing checks are unrelated to your PR directly, merging.

PR #348, opened 2015-12-12
Fix var-dumper suggestion string, fixup 6a1c86bb

denis-sokolov:
https://getcomposer.org/doc/04-schema.md#suggest


denis-sokolov:
@staabm, please confirm this makes sense to you and whether you'd like to reword. This was your added package.

staabm:
@denis-sokolov lgtm thx

PR #350, opened 2015-12-13
Update php-min version to 5.5

staabm:
I would like to propose uplifting the php min version for this major release.
Php 5.5 is the oldest supported php version


staabm:
In case we agree on landing this we could scan the src for code/workarounds handling old phpversions

denis-sokolov:
Fair.

PR #357, opened 2015-12-20
Move utility functions to the utility class, fix #351

### LINKED ISSUE #351: Refactor Misc* methods into functions.php
From https://github.com/filp/whoops/blob/master/src/Whoops/Util/Misc.php
To https://github.com/filp/whoops/blob/master/src/Whoops/functions.php

The Misc class only contains static helper methods.. Those could -for consistency - live with the other helper fns directly in the whoops-namespace

denis-sokolov:


staabm:
:+1:

GrahamCampbell:
:+1:

PR #360, opened 2016-01-03
Highlight the current line, fix #359

### LINKED ISSUE #359: highlight Current Line
i'm using 2.0.0-alpha2 release and highlighting the error line doesn't work

denis-sokolov:
screen shot 2016-01-03 at 20 22 20

### DIFF
--- a/src/Whoops/Resources/views/frame_code.html.php
+++ b/src/Whoops/Resources/views/frame_code.html.php
@@ -29,6 +29,10 @@
             $start = key($range) + 1;
             $code  = join("\n", $range);
         ?>
+            <style>#frame-code-0 .linenums li:nth-child(<?= $line-$start+1 ?>) {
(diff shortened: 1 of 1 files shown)


src/Whoops/Resources/views/frame_code.html.php
```diff
@@ -29,6 +29,10 @@
             $start = key($range) + 1;
             $code  = join("\n", $range);
         ?>
+            <style>#frame-code-0 .linenums li:nth-child(<?= $line-$start+1 ?>) {
```
staabm: Shouldnt we use a css class instead?

denis-sokolov: We can't, this HTML is generated by the prettifier.

staabm:
:+1:

PR #364, opened 2016-01-13
Remove the final restriction on Whoops\Run

guiwoda:
Allows other libraries, frameworks and projects to mock `Whoops\Run` by removing its `final` restriction.

If extending the class is *really* not recommended, I'd favor a big comment on the `README` instead of the `final` keyword.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -14,7 +14,7 @@ use Whoops\Handler\Handler;
 use Whoops\Handler\HandlerInterface;
 use Whoops\Util\Misc;
 
-final class Run
+class Run
 {
     const EXCEPTION_HANDLER = "handleException";
     const ERROR_HANDLER     = "handleError";


mahagr:
:+1:

staabm:
I get the feeling we discussed this in another issue not long ago...

GrahamCampbell:
:-1:

prisis:
:-1:

guiwoda:
@GrahamCampbell @prisis care to elaborate?

guiwoda:
@staabm we are building adapters over multiple services, including error handlers / formatters. Whoops is, for us, one of many error formatting strategies. Our Whoops wrapper cannot be tested because of the `final` restriction on the `Whoops\Run` class, and we can't use the real class in our tests without a lot of heavy "monkey patching" (ob_* functions, echo statements).

denis-sokolov:
Perhaps you want to use [`PrettyPageHandler`](https://github.com/filp/whoops/blob/master/src/Whoops/Handler/PrettyPageHandler.php) directly?

guiwoda:
@denis-sokolov and copy all the rest of the `Run` logic to my side? Why would I do that?

denis-sokolov:
You said ‘error formatting strategies’. Run does none of that, does it?

prisis:
No it doesn't.

@guiwoda because Run should only be used to add handlers.
And the handler will do the rest for you, If you're building some adapters.

guiwoda:
@denis-sokolov I may have been careless on my definition there :sweat_smile: 

Let's call them "error handlers, formatters and output emitters".

In any case, I don't want my adapter to have duplicated code with `Whoops`, it will complicate its use and may even complicate its interface, when the only thing we need is to let `Whoops` do its thing. We may even provide internal access to `Whoops` so that client code can customize it even further (a sort of leaky abstraction). Copying code to our side would complicate all of this.

This is for testability of any code that uses `Whoops\Run`. Right now, if you use `Whoops\Run` in your project / library, chances are that part of your code is not tested. What's the harm on allowing it? Is the inability to extend that much of a restriction?

@prisis right now `Run` has handlers, sets itself on it when it delegates the handling of the error, also has an `inspector` and output logic.

guiwoda:
Anyway, I'd rather talk about the pros / cons of the `final` keyword there, and I can invite anyone interested on the project that needs it to our slack channel, so we discuss design there!

denis-sokolov:
If we are only to speak about `final`, see #330. General purpose is to express use cases that we do not support in code to ensure we have fewer things to break when we change the implementation of Run.
In most cases it is better to compose objects than to extend.

You are right that with Run it’s trickier because of system functions like, heh, `exit`. Since what you’re doing for testing is workarounds already, you can use the following workaround:

```php
namespace Whoops;
function register_shutdown_function($f){
  // Do not actually register a shutdown function
}
``` 

I think this should work.

denis-sokolov:
And if that does not help, we will design a different way for you to hook in and override Whoops behavior.

Key word being ‘design’. Unlike the previous scenario of ‘we chose to overwrite some Run functions, but Whoops authors knew nothing about it, so when Whoops authors changed things around in Run, our code broke’, if we design and provide you with a documented way to resolve your problem, we will know about that use case and will try to not break it in the future.

danives3d:
I've had the same issue; we were extending the Run class so that we could capture the output from the 'PrettyPageHandler' class so we could log each instance where a warning / notice was thrown (as opposed to Whoops fully stopping the page from loading), to then display in a table where the content could be loaded up issue by issue in iFrames (inside of which Zepto is throwing an JS error preventing line navigation but thats a different issue).

I've managed to bypass the problem by taking the methods of the Run class that I require, implementing in a new class and changing some of the methods accordingly so it isn't dependant on the final class. However in trying to use the 'PrettyPageHandler' I still have to provide it with an instance of Whoops, which I'm just creating at the time:

```
$this->handler = new PrettyPageHandler();
...
$this->handler->setRun(new \Whoops\Run());
$this->handler->setInspector($inspector);
$this->handler->setException($exception);
```

So I've managed to bypass the problem, but it does mean if anything changes within Run and elsewhere I may have to readjust the implementation accordingly. I'd be for removing the `final` declaration, or the addition of some extra methods / documentation to allow for overriding.

guiwoda:
@denis-sokolov I'll reead #330! I do agree on the premise that restriction on how a library must be consumed is an overall Good Thing :tm: 

If `Run` would really delegate all its functionality to other objects, and would be designed with dependency inversion in mind (maybe using the template method pattern?), then being final wouldn't be a problem. I'd happily instantiate it in my tests, and provide mocks for its internals.

Or, provide an `RunInterface` that I can mock, then make `Run` your only implementation of it. If your idea is for `Run` to have a small, consistent API, then this shouldn't be a problem.

guiwoda:
@denis-sokolov I've read #330. No one mentioned mocks in the possible scenarios where you'd need to extend `Run`, so I'll guess no one thought about it.

Funny that @mahagr is the only one that actively :+1:'d here, and was the only one that was affected by #330 before!

prisis:
Im :+1: for a RunInterface.

denis-sokolov:
@danives3d, it seems like we would all benefit if instead an option to gather warnings along the way pas in Whoops core! Kindly consider contributing it, and then the onus of maintaining the feature would be on Whoops maintainers.

@guiwoda, your statement about delegating all functionality to a helper is what I like a lot. A `Whoops\PHPRuntimeFacade`, that would have methods for all system functions that we use - `register_shutdown`, `set_error_reporting`, `exit`. Then you can provide another implementation of that as an injected dependency, great. In addition, this would improve our `Run` tests, as currently we face the same issue ourselves - can't test `Run`, cause it has an effect on the global state.
If you agree on such a delegate, are you interested in this topic enough to attempt a PR?

denis-sokolov:
With #374, this is not needed anymore. Thank you for your contribution!

PR #365, opened 2016-01-13
Provide an interface to the `Whoops\Run` class

guiwoda:
As an alternative to #364, this PR allows consumers to hint the `Whoops\RunInterface` in their code, thus allowing all dependent code to be testable without affecting the `final` restriction on `Whoops\Run`.


guiwoda:
Hold on a sec... amending...

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
prisis: Please remove.

guiwoda:
@prisis sorry about that, missed returns hinting on the concrete class, then added some .idea files to the commit on the rush...

GrahamCampbell:
:-1:

guiwoda:
I love you @GrahamCampbell :heart:

prisis:
He really likes that icon haha

guiwoda:
Let's hear what @denis-sokolov thinks about it then :smile:

GrahamCampbell:
As far as I'm concerned there is no way you should ever me trying to mock or extend this class. That's why it was set as final.

guiwoda:
@GrahamCampbell Well then you clearly haven't seen my use case. But this PR is not about extending `Run`, is about adding an interface to it, so that we can replace it in ~testing~ any other context. This could also mean more flexibility on dependencies when someone intends to decorate `Run`, which is the currently recommended extension point.

denis-sokolov:
Thank you for your PR, it is appreciated. It is tidy and seems to achieve the goal stated in the PR description. For the time being, I am not fully convinced an interface would be beneficial for Whoops, especially when an injected mocking delegate is still an option. Until we rule out other options, I will close this. But if we find no other way, we will surely return to this. Thanks again.

guiwoda:
@denis-sokolov I'm sorry to hear that.
What do you mean by an "injected mocking delegate"? Something as we're discussing on #364?

denis-sokolov:
@guiwoda, yes, I tried to use terminology that you introduced. Here:

```php
interface Run\SystemFacade {
  public function exit();
  public function set_error_handler($callback);
  public function restore_exception_handler();
}


class Run {
  public function __construct(Run\SystemFacade $system = null){
    $this->system = $system ?: $defaultSystem;
  }

  public function unregister(){
    $this->system->restore_exception_handler();
    $this->system->restore_error_handler();
  }
}
```

guiwoda:
@denis-sokolov I started it yesterday, I'll see if I can PR by the weekend

PR #374, opened 2016-01-22
System "runtime" facade implementation

guiwoda:
As discussed in #364 and #365, a system facade is implemented to allow consumers of Whoops\Run to mock any system call that it would do through native functions.


denis-sokolov:
Thanks, I think this is a great improvement!

PR #378, opened 2016-02-01
Issue 367 svg

prisis:


denis-sokolov:
That’s awesome!
Could you move the license disclaimers from CSS to HTML, to go along with the image data?

prisis:
done.

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -490,18 +490,15 @@ pre.sf-dump {
...
+      .search-for-help li a svg.google {
+        fill: #fff;
+      }
+      .search-for-help li a svg.google path {
+        background-size: contain;
+      }
+      .search-for-help li a svg.stackoverflow {
+        fill: #fff;
+      }
+      .search-for-help li a svg.google path {
```
denis-sokolov: This is a duplicate selector and a duplicate body. 6 lines above there’s the same.

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -490,18 +490,15 @@ pre.sf-dump {
...
+      .search-for-help li a svg.google {
+        fill: #fff;
+      }
+      .search-for-help li a svg.google path {
+        background-size: contain;
+      }
+      .search-for-help li a svg.stackoverflow {
```
denis-sokolov: This can be combined with the selector 6 lines above it quite nicely:

```css
.search-for-help li a svg {
  fill: #fff;
}
```

src/Whoops/Resources/views/header.html.php
```diff
@@ -20,8 +20,26 @@
...
-      <li><a target="_blank" href="https://stackoverflow.com/search?q=<?php echo urlencode(implode('\\', $name).' '.$message) ?>" title="Search for help on Stack Overflow."><i class="stackoverflow"></i></a></li>
+      <li>
+        <a target="_blank" href="https://google.com/search?q=<?php echo urlencode(implode('\\', $name).' '.$message) ?>" title="Search for help on Google.">
+          <!-- Google icon by Alfredo H, from https://www.iconfinder.com/alfredoh -->
+          <!-- Creative Commons (Attribution 3.0 Unported) -->
+          <!-- http://creativecommons.org/licenses/by/3.0/ -->
+          <svg class="google" height="16" viewBox="0 0 512 512" width="16" xmlns="http://www.w3.org/2000/svg">
+            <path d="M457.732 216.625c2.628 14.04 4.063 28.743 4.063 44.098C461.795 380.688 381.48 466 260.205 466c-116.024 0-210-93.977-210-210s93.976-210 210-210c56.703 0 104.076 20.867 140.44 54.73l-59.205 59.197v-.135c-22.046-21.002-50-31.762-81.236-31.762-69.297 0-125.604 58.537-125.604 127.84 0 69.29 56.306 127.97 125.604 127.97 62.87 0 105.653-35.966 114.46-85.313h-114.46v-81.902h197.528z"/>
+            </svg>
```
denis-sokolov: Tiny indentation inconsistency. `</svg>` should be outdented once.

denis-sokolov:
Thank you! Three other tiny fixes and I’m merging this in.

vinkla:
@prisis maybe you can run the svgs trough https://jakearchibald.github.io/svgomg/

prisis:
@vinkla  i did :D

vinkla:
Okay, good. Though, we could bring it down even more.

screen shot 2016-02-01 at 11 36 46

denis-sokolov:
Please let’s not save bytes here. We really don’t care about a 100kb here and there.

prisis:
@denis-sokolov ready to merge

denis-sokolov:
Great job! :+1:

vinkla:
@denis-sokolov what? Of course we do. Why wouldn't we?

vinkla:
@prisis Good work :+1:

staabm:
thanks guys

denis-sokolov:
@vinkla We briefly [discussed this before](https://github.com/filp/whoops/pull/128#issuecomment-35195358). To re-state my opinion: the PrettyPageHandler is used in development, so no impact for production. It only shows up for errors, which should not be a developer’s constant experience, and once in a while a tiny delay matters not much. A developer is usually developing on a server that’s localhost or very close to that, meaning you’d need tens of megabytes of content to make them notice a delay.

vinkla:
Yeah, that is true. Though, we should strive to optimize as much as possible even when it is used for local development. Since this is a package which is used by many we should stand as an example of good practice. The extra bytes will probably not matter for the single developer but imagine how much it would be after a couple of thousands downloads with composer.

IMO opinion we should even compile all styles and scripts into minified single files for better performance.

denis-sokolov:
@vinkla, this would mean that contributing to Whoops is harder, as someone will need to maintain those minified files. This would mean that Whoops would get even less attention both from me and possibly from other maintainers. Everything’s a budget trade-off in this world. =)

vinkla:
Yeah, maybe. Let me know if I can help you guys with that. If it is setup correctly it would simple be a single command to run `npm install` which would install and build all assets automagically.

denis-sokolov:
But we certainly don’t want to have our compiled artifacts to be in the source tree, as that would mean building with every commit, and big useless diffs in PRs.

But the Packagist packs Whoops for distribution over Composer taking source code from a commit I tag. Setting extra stuff for it to do is harder.

Also, how do you handle the ability to develop Whoops without recompiling assets and without a fancy setup?

All of that can be solved, but if someone has so much time they are willing to put into Whoops, there are more important and useful improvements that can be done. :)

vinkla:
The assets would only have to be compiled when there are actual updates done to either the CSS or JavaScript. I agree it wouldn't be very nice to keep the compiled assets within the source tree but we already do this today and it is quite common. This would also allow us to not keep copies of other libraries like Zepto within our source code. Instead we would only have a `whoops.js` file.

denis-sokolov:
Which would imply that every PR that does changes to CSS or JS would have a much bigger and more confusing diff than needed. Also this would require all the contributors to build themselves before committing, or me to build for them before merging. Some projects decide that it’s a good idea, this project has decided that it isn’t.

Thank you for your insightful comments here and I enjoyed this talk and it was beneficial and thoughtful.
Please do not take it personally when I tell this is not a discussion and Whoops source will not include a build step for JS and CSS for the time being.

vinkla:
Okay, lets leave it like it is.

staabm:
I agree with @denis-sokolov that this kind of change would make more harm then good. bytesize is not important for this project.

PR #384, opened 2016-02-13
Added dumping of variables under code

damnedest:
This is my implementation of this ticket. Since in some cases in left side could be a lot of stuff I draw only types there.

http://imgur.com/xOkkyfI

Related to #337

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -136,6 +136,22 @@ header {
       font-size: 14px;
     }
 
+    .frame-arg {
+      margin-left: 20px;
+      display: block;
+    }
+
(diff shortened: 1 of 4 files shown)


staabm:
No sure what you suggest here. The col at the left shows a type and its value, so its not different from #337?

damnedest:
On left column we show only types, on right column under code we show full information about variables

staabm:
I cant follow.
On the left I can see `string(16) 'Something broke'` - which is a type and its value.
In the main column you can see the string `Something broke` - in a place where it is hard to associate the string to be an argument...

Your example should show how things will look when a object or array is passed: the case with a simple string does not represent the worst case.

damnedest:
Close in case of #387

PR #386, opened 2016-02-18
Added Sublime Text 3 "subl" protocol support

doublehelix:
Updated to support the protocol for the SUBL PROTOCOL package.
This particular protocol must use the raw (unencoded) file path, as well as *nix style line endings (even on Windows).
see: https://packagecontrol.io/packages/subl%20protocol


denis-sokolov:
Thank you for the contribution.

Unfortunately, I can not accept it at this time. The package in question is far from being popular for Sublime 3. We don't want to have predefined values for all possible plugins people post on the web.

Whoops users can manually add specific editors:

```php
$handler->setEditor(function($file, $line) {
    return "subl://$file:$line";
});
```

PR #387, opened 2016-02-18
Dump frame arguments below the source code

jonasdt:
This builds on top of #337, but dumps the arguments below the source code (similar to #384).

Collapsed by default:
http://imgur.com/9augmq7
Expand second argument:
http://imgur.com/oXfrSfa
Expand first and second argument:
http://imgur.com/JJQ1oF0
No arguments:
http://imgur.com/uWVch0e

### DIFF
--- a/src/Whoops/Resources/js/whoops.base.js
+++ b/src/Whoops/Resources/js/whoops.base.js
@@ -119,4 +119,10 @@ Zepto(function($) {
     e.preventDefault();
     $.get(this.href);
   });
+
+  // Symfony VarDumper: Close the by default expanded objects
+  $('.sf-dump-expanded')
+    .removeClass('sf-dump-expanded')
+    .addClass('sf-dump-compact');
+  $('.sf-dump-toggle span').html('&#9654;');
 });
--- a/src/Whoops/Util/HtmlDumperOutput.php
+++ b/src/Whoops/Util/HtmlDumperOutput.php
@@ -6,6 +6,11 @@
 
 namespace Whoops\Util;
 
+/**
+ * Used as output callable for Symfony\Component\VarDumper\Dumper\HtmlDumper::dump()
+ *
+ * @see TemplateHelper::dump()
+ */
 class HtmlDumperOutput
 {
     private $output;
--- a/src/Whoops/Util/HtmlDumperOutput.php
+++ b/src/Whoops/Util/HtmlDumperOutput.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Whoops - php errors for cool kids
+ * @author Filipe Dobreira <http://github.com/filp>
+ */
+
+namespace Whoops\Util;
+
+class HtmlDumperOutput
+{
+    private $output;
+
+    public function __invoke($line, $depth)
+    {
+        // A negative depth means "end of dump"
+        if ($depth >= 0) {
+            // Adds a two spaces indentation to the line
+            $this->output .= str_repeat('  ', $depth) . $line . "\n";
+        }
+    }
+
+    public function getOutput()
+    {
+        return $this->output;
+    }
+
+    public function clear()
+    {
+        $this->output = null;
+    }
+
+}
--- a/src/Whoops/Util/TemplateHelper.php
+++ b/src/Whoops/Util/TemplateHelper.php
@@ -28,6 +28,11 @@ class TemplateHelper
     private $htmlDumper;
 
     /**
+     * @var HtmlDumperOutput
+     */
+    private $htmlDumperOutput;
+
+    /**
      * Escapes a string for output in an HTML document
      *
      * @param  string $raw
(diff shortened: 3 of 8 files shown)


prisis:
:+1: can you squash it? :)

src/Whoops/Resources/js/whoops.base.js
```diff
@@ -119,4 +119,10 @@ Zepto(function($) {
     e.preventDefault();
     $.get(this.href);
   });
+
+  // Symfony VarDumper: Close the by default expanded objects
+  $('.sf-dump-expanded')
+    .removeClass("sf-dump-expanded")
+    .addClass("sf-dump-compact");
+  $(".sf-dump-toggle span").html('▶');
```
staabm: please use a html sequence instead of a UTF8 char (to make it work on non utf8 websites)

src/Whoops/Util/TemplateHelper.php
```diff
@@ -22,6 +22,16 @@ class TemplateHelper
     private $variables = array();
 
     /**
+     * @var HtmlDumper
+     */
+    private $htmlDumper;
```
staabm: I would prefer the static vars in `getDumper` as it read before

src/Whoops/Util/HtmlDumperOutput.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * Whoops - php errors for cool kids
+ * @author Filipe Dobreira <http://github.com/filp>
+ */
+
+namespace Whoops\Util;
+
+class HtmlDumperOutput
```
staabm: please add a class comment what the use case for this class is

staabm:
really like it. besides the few nits I commented above it looks like a better alternative to my initial PR

src/Whoops/Util/HtmlDumperOutput.php
jonasdt: done

src/Whoops/Util/TemplateHelper.php
jonasdt: Can you explain why this is better? The TemplateHelper instance is reused during the render step, so it's not necessary to make it a static variable. And in case you wanna render multiple pages, we should have the javascript on each page (which will happen now because we create a new instance in PrettyPageHandler::handle()).

staabm: hmm usage of the helper on different pages is actually a good argument. I guess having the properties will also ease unit testing because we dont depend on a semi-global state.

agree, leave it as is.

jonasdt:
@staabm I've made the changes (except for the static variable thing). I can squash the commits if it is required.

staabm:
:+

shadowhand:
Please merge this! It is really nice to have arguments, especially when working with templates where multiple variables might be passed through the same functions. Determining what variable is causing an issue is extremely hard when no arguments are visible.

damnedest:
Looks great

damnedest:
Any news about merging this request?

grevus:
+1

denis-sokolov:
I will get to it, sorry for the delay.

damnedest:
Any updates?

grevus:
+

serafimovich:
+

vvolynets:
+

prisis:
Guys wait a bit, maybe @denis-sokolov has a bit to do the last past days

denis-sokolov:
Seems good, thank you for your contributions, @staabm, @jonasdt!

kktsvetkov:
@denis-sokolov @staabm @jonasdt Doesn't this "dump frame arguments" change have some security implications? Just like $_SERVER and $_ENV, stack frame arguments might contain security-sensitive data, such as credentials or API keys and tokens. Perhaps the blacklist feature must expand to cover arguments as well ?

Here's a somewhat real example: imagine a method/function that takes credentials as arguments and uses them in a Guzzle request, but the request fails and an exception is thrown; the exception message itself might not contain any sensitive data (although, sometimes they do - mainly tokens and API keys), but when you start looking at the backtrace,  you are going to see the credentials as the arguments in the stack frame.

jonasdt:
It is not important for our workflow as we only use Whoops in development, but I understand your point that some people might forget to disable it on production. The same can be said about `phpinfo` ...

staabm:
Running it in production is not a valid use-case. Nevertheless I feel a blacklist for some sensitive parameters makes sense though

kktsvetkov:
I thought about this more today and there are several places where sensitive informations can be exposed. It is not just the data tables and the arguments dump, but sometimes also the exception messages, and the source code chunks. Blacklist needs more wholistic approach, but even if there is such a solution, I agree that this is all meant to be used for development only. 

I've posted another idea I have about how to "hide" Whoops on non-development environments at [604](https://github.com/filp/whoops/issues/604)

PR #391, opened 2016-03-12
Add RunInterface

### LINKED ISSUE #390: Add interface for Run
The class `Whoops\Run` is declared as final. While it may be a good choice, having this class as final without any interface just makes it impossible to mock and to test.

So something as simple as:

```php
class Dummy
{
    /**
     * @var Run
     */
    private $run;

    public function __construct(Run $run = null)
    {
        $this->run = ($run === null) ? new Run() : $run;
    }

    /**
     * {@inheritdoc}
     */
    public function render(Request $request, \Exception $exception)
    {
        $this->run->pushHandler(new PrettyPageHandler());

        if ($request->isJson()) {
            // Since this handler is first in the stack, it will be executed before the error page handler, and will
            // have a chance to decide if anything needs to be done.
            $this->run->pushHandler(new JsonResponseHandler);
        }

        return new Response(
            $this->run->handleException($exception)
        );
    }
}
```

cannot be ever unit tested properly...

theofidry:
Proposed changes:

- Add an interface `Whoops\RunInterface` and make `Whoops\Run` implements this interface
- Move `Run` constants in `RunInterface` (is BC compatible)
- Made `Run` `protected` methods/properties `private` as `Run` cannot be inherited (should I move that in a separate PR?)

Closes #390.


tests/Whoops/RunTest.php
```diff
@@ -15,6 +15,11 @@
 
 class RunTest extends TestCase
 {
+    public function testImplementsRunInterface()
+    {
+        class_implements('Whoops\\Run', 'Whoops\\RunInterface');
```
denis-sokolov: I think you need an `assert` here. Right now this tests nothing.

denis-sokolov:
> Made Run protected methods/properties private as Run cannot be inherited (should I move that in a separate PR?)

Generally, yes, a separate PR for such unrelated things would be preferred. But in this specific case now that we already have it, don’t bother.

src/Whoops/Run.php
```diff
@@ -15,29 +15,25 @@
 use Whoops\Util\Misc;
 use Whoops\Util\SystemFacade;
 
-final class Run
+final class Run implements RunInterface
 {
-    const EXCEPTION_HANDLER = "handleException";
```
staabm: Are the consts still accessible via `Run::` from the outside?

src/Whoops/RunInterface.php
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return bool
+     * @throws ErrorException
+     */
+    public function handleError($level, $message, $file = null, $line = null);
+
+    /**
+     * Special case to deal with Fatal errors and the like.
+     */
+    public function handleShutdown();
```
staabm: Isnt this method a impl-detail and should therefore not be part of the interface?

src/Whoops/Run.php
theofidry: Yep they are, but I'll add a test just in case :)

tests/Whoops/RunTest.php
theofidry: indeed, my bad...

src/Whoops/RunInterface.php
theofidry: well to be honest I just took all of `Run` public functions to the interface. @denis-sokolov will know better than me if we should or not :)

theofidry:
Updated :)

src/Whoops/RunInterface.php
denis-sokolov: I was thinking about this a bit too long. Let's keep it in the interface.

denis-sokolov:
Thank you!

PR #392, opened 2016-03-12
Cache composer packages and secure PHP versions

theofidry:
For PHP version quotes, it looks like it was a mistake on the Travis team side not to include them in the doc (they changed it now), as without them, versions are interpreted as numbers instead of strings which can cause some issues (it did for some versions of nodejs). For the whole discussion thread, you can check https://github.com/api-platform/core/pull/449#discussion_r55392343.

The other change is simply installing the dist packages (lighter, so faster) and caching them to speed up the `composer install`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
 language: php
 
 php:
-  - 5.5.9
-  - 5.5
-  - 5.6
-  - 7.0
+  - '5.5.9'
+  - '5.5'
+  - '5.6'
+  - '7.0'
   - hhvm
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Could you please `git commit --amend` and `git push --force`? Travis did not trigger for this PR, for whatever reason, and I can not or do not know how to manually start a Travis run for a PR.

theofidry:
Here you go :)

denis-sokolov:
Thank you!

Issue #215, opened 2014-06-10
Replay monolog errors in whoops


garygreen:
Is it possible to 'replay' errors generated in monolog format (i.e. file or JSON format) and re-generate a whoops error screen? I have a list of log files that are in Laravel's default monolog format and wonder how difficult it would be to create a log viewer that shows the exceptions in whoops -- is this currently possible?

denis-sokolov:
Closing due to inactivity. If anyone wants to work on this, feel free to speak up.

PR #394, opened 2016-03-14
Custom inspector

jonasdt:
I've made the following backwards compatible changes:
- The bundled handlers only need and use the inspector, not the exception. This was discussed in #215.
- Allow subclasses of Inspector.
- Allow using your own Inspector in Run.

I need this because in production I'm storing [flattened](http://api.symfony.com/2.7/Symfony/Component/HttpKernel/Exception/FlattenException.html) exceptions in the database and I'm using my own inspector (for flattened exceptions) to generate the same pretty page as in development.

### DIFF
--- a/src/Whoops/Exception/Inspector.php
+++ b/src/Whoops/Exception/Inspector.php
@@ -81,6 +81,36 @@ class Inspector
         return $this->exception->getMessage();
     }
 
+    /**
+     * @return int
+     */
+    public function getCode()
+    {
+        if ($this->exception instanceof \ErrorException) {
+            // ErrorExceptions wrap the php-error types within the "severity" property
+            return Misc::translateErrorCode($this->exception->getSeverity());
+        }
+
+        return $this->exception->getCode();
+    }
+
+    /**
+     * @return string
+     */
+    public function getFile()
+    {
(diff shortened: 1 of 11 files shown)


denis-sokolov:
This seems interesting, but a bit too complicated for me right now. I will get back to it later.

src/Whoops/Exception/Inspector.php
```diff
@@ -42,7 +42,11 @@ public function getException()
     }
 
     /**
+     * Use getClass.
+     *
      * @return string
+     *
+     * @deprecated
```
denis-sokolov: When you deprecate a method you want to replace its body to `$this->newMethod()`, so that if we change the new method, the deprecated would include the new functionality.

garygreen:
Yes! This will make it a lot easier to replay the errors in whoops. @jonasdt are you releasing your log viewer as tool? (please)? :smile:

mahagr:
Interesting, +1 from me.

denis-sokolov:
I’m sorry I’ve left this PR hanging. It’s a lot and I don’t have that bandwidth for that. Since it’s now been a while, I will close it as outdated. Please speak up if you want to refresh this going forward.

PR #395, opened 2016-03-14
Overwrite globals and canSendheaders

jonasdt:
I've made 2 changes:
- Allow to manually set the globals in PrettyPageHandler. This way I can rebuild the pretty page from captured data (e.g. in database).
- Allow to manually set if the handler is allowed to send headers. This allows for better integration with a framework. E.g. I capture the output of the handlers and make a Symfony response from this.

### DIFF
--- a/src/Whoops/Handler/Handler.php
+++ b/src/Whoops/Handler/Handler.php
@@ -8,6 +8,7 @@ namespace Whoops\Handler;
 
 use Whoops\Exception\Inspector;
 use Whoops\Run;
+use Whoops\Util\Misc;
 
 /**
  * Abstract implementation of a Handler.
@@ -37,6 +38,11 @@ abstract class Handler implements HandlerInterface
      * @var \Throwable $exception
      */
     private $exception;
+
+    /**
+     * @var bool
+     */
+    private $sendHeaders;
 
     /**
      * @param Run $run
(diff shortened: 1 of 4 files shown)


denis-sokolov:
I'm not sure it makes sense to customize global variables. Perhaps, instead, we can simply allow disabling global output, and then you can display whatever you want through `extraTables` feature?

denis-sokolov:
I also do not agree to add `canSendHeaders` to handlers. Instead, make `PrettyPageHandler` expose a `generateOutput` method that you can use programatically, and keep `handle` doing `header`, `echo` and `exit`.

jonasdt:
> I'm not sure it makes sense to customize global variables. Perhaps, instead, we can simply allow disabling global output, and then you can display whatever you want through extraTables feature?

That will also work as whoops isn't doing anything with the globals other then dumping them.
Currently, I'm using it to set the session data as Laravel isn't using $_SESSION. I also use it rebuild the page for exceptions that happen in production.

Should I make a new pull request?

jonasdt:
> I also do not agree to add canSendHeaders to handlers. Instead, make PrettyPageHandler expose a generateOutput method that you can use programatically, and keep handle doing header, echo and exit.

That's a bit annoying because then I cannot use Run to handle the exception, and hence also loose the output buffering and support for multiple handlers. Also, currently the handlers receive a Run instance.

It would be better if the handlers return an array (or an object if you want) that describes the result of handling the exception (output, headers, exit), but that's a bigger backwards incompatible change.

staabm:
> That will also work as whoops isn't doing anything with the globals other then dumping them.
> Currently, I'm using it to set the session data as Laravel isn't using $_SESSION.

this should be a separate PR

sagikazarmark:
Although I don't really agree with the approach, I fully agree with the intent.

One perfect example is the case of environment variables: they often contain sensitive information which could leak if we don't specifically prevent them from appearing. I know the pretty page handler should only be used in development environment, but one small mistake could cause it to be published to everyone. One could argue though if it is more important to see these credentials during development.

denis-sokolov:
Since it’s now been a while, I will close it as outdated. Please speak up if you want to refresh this going forward.

PR #396, opened 2016-03-14
Remove all internals for objects not in the application namespaces

jonasdt:
Dumping the frame arguments is very useful, but can make the response very big. For instance, in a vanilla Laravel setup the page becomes 14.5mb. 14mb comes from dumping Laravel internals (like the Router) and Closures that use those objects. This information isn't useful, so it's ok to hide it.

This change will optionally (if PrettyPageHandler::$applicationNamespaces are defined) remove all internals for objects that are not defined in the application (e.g. framework objects). For Laravel, this brings the response for Laravel back to 300-400k. This can be further reduced by cutting all internals of Closures (now it will still show some basic information), but I assume this little overhead is probably ok.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -8,6 +8,8 @@ namespace Whoops\Handler;
 
 use InvalidArgumentException;
 use RuntimeException;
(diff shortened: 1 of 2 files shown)


denis-sokolov:
We have merged a different implementation of yours in #404. Let us know if this is still relevant.

PR #397, opened 2016-03-14
Break filenames and namespaces on the delimiter

jonasdt:
Before:
http://imgur.com/nz2NMSq

After:
http://imgur.com/Bof5azS

### DIFF
--- a/src/Whoops/Exception/Frame.php
+++ b/src/Whoops/Exception/Frame.php
@@ -58,7 +58,9 @@ class Frame implements Serializable
         if ($shortened && is_string($file)) {
             // Replace the part of the path that all frames have in common, and add 'soft hyphens' for smoother line-breaks.
             $dirname = dirname(dirname(dirname(dirname(dirname(dirname(__DIR__))))));
-            $file = str_replace($dirname, "&hellip;", $file);
+            if ($dirname !== '/') {
+                $file = str_replace($dirname, "&hellip;", $file);
+            }
             $file = str_replace("/", "/&shy;", $file);
         }
 
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -171,7 +171,6 @@ header {
 
     .frame-file {
       font-family: "Inconsolata", "Fira Mono", "Source Code Pro", Monaco, Consolas, "Lucida Console", monospace;
-      word-wrap: break-word;
       color: #a29d9d;
     }
 
@@ -256,6 +255,10 @@ header {
       font-size: 10px;
       color: white;
     }
+
+.delimiter {
+  display: inline-block;
+}
 
 .data-table-container label {
   font-size: 16px;
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -157,6 +157,14 @@ header {
     }
 
     .frame-index {
+      float: left;
+    }
+
+    .frame-method-info {
+      margin-left: 24px;
+    }
+
+    .frame-index {
       font-size: 11px;
       color: #a29d9d;
       background-color: rgba(0, 0, 0, .05);
--- a/src/Whoops/Util/TemplateHelper.php
+++ b/src/Whoops/Util/TemplateHelper.php
@@ -72,6 +72,39 @@ class TemplateHelper
         );
     }
 
+    /**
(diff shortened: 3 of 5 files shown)


src/Whoops/Exception/Frame.php
```diff
@@ -58,7 +58,9 @@ public function getFile($shortened = false)
         if ($shortened && is_string($file)) {
             // Replace the part of the path that all frames have in common, and add 'soft hyphens' for smoother line-breaks.
             $dirname = dirname(dirname(dirname(dirname(dirname(dirname(__DIR__))))));
-            $file = str_replace($dirname, "&hellip;", $file);
+            if ($dirname !== '/') {
+                $file = str_replace($dirname, "&hellip;", $file);
```
denis-sokolov: Now that `TemplateHelper::shorten` is added, this here should not have repeated code, but should use that instead.

- [ ] Replace this repeated part with a call to `TemplateHelper::shorten`

src/Whoops/Util/TemplateHelper.php
```diff
@@ -72,6 +72,39 @@ public function escapeButPreserveUris($raw)
...
+     * Makes sure that the given string breaks on the delimiter.
+     *
+     * @param  string $delimiter
+     * @param  string $s
+     * @return string
+     */
+    public function breakOnDelimiter($delimiter, $s)
+    {
+        $parts = explode($delimiter, $s);
+        foreach ($parts as &$part) {
```
denis-sokolov: I thought we don't use by-reference for a while now. Does this have some interesting meaning I am missing?

src/Whoops/Util/TemplateHelper.php
```diff
@@ -72,6 +72,39 @@ public function escapeButPreserveUris($raw)
...
     }
 
+    /**
+     * Makes sure that the given string breaks on the delimiter.
+     *
+     * @param  string $delimiter
+     * @param  string $s
+     * @return string
+     */
+    public function breakOnDelimiter($delimiter, $s)
```
denis-sokolov: You can replace the implementation of this function with a `str_replace($s, $delimiter, $delimiter . '&shy;')`.
The DOM is simpler, the PHP is simpler.

- [ ] Replace `div.delimiter` with `&shy`s.

src/Whoops/Resources/css/whoops.base.css
```diff
@@ -157,6 +157,14 @@ header {
     }
 
     .frame-index {
+      float: left;
```
denis-sokolov: floats can always be avoided and replaced with a better `inline-block` or `display: flex`.

In this case, perhaps simply setting `.frame-index, .frame-method-info { display: inline-block }` solves the problem better.

 - [ ] Try inline-blocks instead of floats

denis-sokolov:
Great improvement, @jonasdt, thank you! I have added some comments to take care of, but in terms of the main idea, this is getting totally merged.

src/Whoops/Exception/Frame.php
jonasdt: The shortened argument isn't used by whoops, I left it in case some people made a custom handler and expect that the function returns shortened paths. I don't think the repeated code harms anyone.
Anyway, how should I use the TemplateHelper::shorten from Frame::getFile? Make it a static method, or create a TemplateHelper instance inside getFile?

src/Whoops/Util/TemplateHelper.php
jonasdt: By using a reference php doesn't have to make a copy of the string and you can update the array value without manually updating the array. The following code gives the same result
```php
 foreach ($parts as $k => $part) {
    $parts[$k] = '<div class="delimiter">' . $part . '</div>';
}
```
I used a reference without thinking about it :)

src/Whoops/Util/TemplateHelper.php
jonasdt: No it's not the same. By wrapping the parts with a div, browsers will not break the text inside a part, but put the div on the next line. Also, browsers put a dash after soft hypen.
Compare the pictures in the pull request description.

src/Whoops/Resources/css/whoops.base.css
jonasdt: I think inline-blocks are only going to work if you specify a width for both blocks (I've just tried it). But then you have to do it in percents because the width of the left-panel depends on the screen size.
What's wrong with a float:left here?

PR #398, opened 2016-03-14
Small ui improvements

jonasdt:
For details see commit messages.

Remark regarding bbe394e:  There was a bug in the container css that caused the left panel to disappear sometimes, depending on the content and the size of the window. Sometimes it happened after closing the chrome developers tools.

### DIFF
--- a/src/Whoops/Resources/js/whoops.base.js
+++ b/src/Whoops/Resources/js/whoops.base.js
@@ -21,7 +21,16 @@ Zepto(function($) {
   var $activeLine     = $frameContainer.find('.frame.active');
   var $activeFrame    = $container.find('.frame-code.active');
   var $ajaxEditors    = $('.editor-link[data-ajax]');
-  var headerHeight    = $('header').height();
+  var $header         = $('header');
+
+  $header.on('mouseenter', function () {
+    if ($header.find('.exception').height() >= 145) {
+      $header.addClass('header-expand');
+    }
+  });
+  $header.on('mouseleave', function () {
+    $header.removeClass('header-expand');
+  });
 
   var highlightCurrentLine = function() {
     // Highlight the active and neighboring lines for this frame:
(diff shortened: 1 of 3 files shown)


src/Whoops/Resources/js/whoops.base.js
```diff
@@ -94,20 +104,34 @@ Zepto(function($) {
     return actionMsg;
   }
 
+  function scrollIntoView($node, $parent) {
+    var nodeOffset = $node.offset(),
+        nodeTop = nodeOffset.top,
```
denis-sokolov: Regardless of personal preference, the surrounding code uses one `var` per variable style, so new code should do that, too.

- [ ] Use consistent var style

src/Whoops/Resources/js/whoops.base.js
```diff
@@ -94,20 +104,34 @@ Zepto(function($) {
     return actionMsg;
   }
 
+  function scrollIntoView($node, $parent) {
```
denis-sokolov: Although the commit message is supposed to explain something, I still don't fully understand what this does differently. Could you explain?

- [ ] Clarify custom `scrollIntoView`

denis-sokolov:
Great, including `j/k`. Needs extra testing due to the amount of layout changes, though.

src/Whoops/Resources/js/whoops.base.js
jonasdt: The old code always keeps the active frame at the top. Check the blue line in the active frame, the first time it makes a jump, afterwards it stays on top
https://www.dropbox.com/s/s916mej3hva6fk6/before.mov?dl=0

The new code does what you would expect to do
https://www.dropbox.com/s/tjh95qj52xkwd3u/after.mov?dl=0

src/Whoops/Resources/js/whoops.base.js
jonasdt: Done

denis-sokolov:

PR #399, opened 2016-03-14
Prettify updates

jonasdt:
See commit messages. 

The purpose of eb61451 and 915a1ad is to make the code before and after prettify look the same, the only difference is the syntax highlighting.


GrahamCampbell:
this doesn't look right?

GrahamCampbell:
you've deleted all the async code

jonasdt:
Yup, the file is included in the response now. I don't see a performance difference, and now it works when I'm on the plane :)

src/Whoops/Resources/views/frame_code.html.php
```diff
@@ -29,20 +29,6 @@
             $start = key($range) + 1;
             $code  = join("\n", $range);
         ?>
-            <style>
-             #frame-code-0 .linenums li:nth-child(<?= $line-$start ?>) {
-                 background-color: rgba(255, 100, 100, .07);
```
denis-sokolov: You probably did not mean to also remove background-color highlight from the lines.

 - [ ] Clarify the background-color highlight.

denis-sokolov:
I agree with removing reliance on 3rd party servers.

src/Whoops/Resources/views/frame_code.html.php
jonasdt: The background-color is defined in [whoops.base.css](https://github.com/jonasdt/whoops/blob/915a1ad1762e443280df6f42e35033a0aaa3bea1/src/Whoops/Resources/css/whoops.base.css#L353-L358).

masterwto:
I agree with removing reliance on 3rd party servers, too, as my project is developed and deployed off internet.

masterwto:
Has been talking for almost a whole year!
I think it's time to merge this PR.

denis-sokolov:
Merged in 1414942. Took a bit of work to get this up-to-date, but I kept your authorship, @jonasdt. Thanks for your contribution!

PR #400, opened 2016-03-14
Filter frames belonging to the application.

jonasdt:
Similar to how [better_errors](https://github.com/charliesome/better_errors) and [prone](https://github.com/magnars/prone) do it. This allows hiding irrelevant stack frames (e.g. generated by a framework).
To enable, set `PrettyPageHandler::$applicationPaths `. The shortcut `n` can be used to switch between application and all frames.
The page looks exactly the same as before if no application paths are set.

Screenshot:
http://imgur.com/0E2I8GO

We show all frames if the last frame was generated by the framework.
http://imgur.com/pBIVyNO

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -53,6 +53,11 @@ class PrettyPageHandler extends Handler
     private $pageTitle = "Whoops! There was an error.";
 
     /**
+     * @var array[]
+     */
+    private $applicationPaths;
+
+    /**
      * A string identifier for a known IDE/text editor, or a closure
      * that resolves a string that can be used to open a given file
      * in an editor. If the string contains the special substrings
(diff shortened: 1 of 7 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -137,6 +142,19 @@ public function handle()
             $code = Misc::translateErrorCode($inspector->getException()->getSeverity());
         }
 
+        // Detect frames that belong to the application.
+        if ($this->applicationPaths) {
+            /* @var \Whoops\Exception\Frame $frame */
+            foreach ($frames as $frame) {
+                foreach ($this->applicationPaths as $path) {
+                    if (substr($frame->getFile(), 0, strlen($path)) === $path) {
```
staabm: Just thinking loud: wouldnt it make more sense to filter based on class namespaces?

denis-sokolov: It seems slightly more practical to do paths. You usually know your application code is in `whatever/src`. And then whether its simple functions or plain files, incorrect namespaces, it will all work here.

denis-sokolov:

Issue #401, opened 2016-03-14
Version 2.1.0 takes a very long time to render errors


patroniton:
I'm using Laravel 5.2 and when I call `return parent::render($request, $e);` in the `render` method in `App\Exceptions\Handler.php` it takes a very long time (2-3 minutes) to display the error page. My CPU usage also spikes to 70-80% while it's running.

I'm also using Graham's Laravel Exceptions package [here](https://github.com/GrahamCampbell/Laravel-Exceptions).

If I catch my own exception or return from the method without calling `return parent::render($request, $e);` then the page renders as expected.

This doesn't happen in version 2.0.0, so for now I'm going to keep using that. If you need any more information from me on this issue I'll try my best to provide you with it.

GrahamCampbell:
I assume this is because whoops 2.1 is trying to dump some heavy ass objects, and that's taking ages.

pelim:
same issue 100% cpu

staabm:
could someone create a blackfire.io profile for this slow case?

denis-sokolov:
@jonasdt, would you please take a look at this, and whether it is related to #387?

jonasdt:
Check #396, for Laravel you can use something like

```php
$handler = new PrettyPageHandler();
$handler->setApplicationNamespaces(['App\\']);
```

It probably makes sense to disable dumping arguments by default.

GrahamCampbell:
>  for Laravel you can use something like

That only works if you dump all your code in your application, but does not work if you have a haveily componentized system.

GrahamCampbell:
> It probably makes sense to disable dumping arguments by default.

Agreed.

jonasdt:
> That only works if you dump all your code in your application, but does not work if you have a heavily componentized system.

If you don't have a common namespace, then you can do something like

```php
$handler->setApplicationNamespaces(['Component1\\', 'Component2\\', 'Component3\\', ...]);
```

jonasdt:
Instead of disabling dumping of arguments entirely, we can also cut all internals of the dumped objects (i.e. only show the classname) by default.

@patroniton and @pelim 
Can you try my pull request #402? You can include the following lines in your composer.json file

```json
    "repositories": [{
        "type": "vcs",
        "url": "https://github.com/jonasdt/whoops"
    }],
    "require": {
        "filp/whoops": "dev-cut-internals"
    },
```

I've tried it for some of my Laravel projects and performance is fine.

rossbearman:
[...]

PR #402, opened 2016-03-16
Don't dump object internals

jonasdt:
See discussion in #401

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -8,6 +8,7 @@ namespace Whoops\Handler;
 
 use InvalidArgumentException;
 use RuntimeException;
+use Symfony\Component\VarDumper\Cloner\VarCloner;
 use UnexpectedValueException;
 use Whoops\Exception\Formatter;
 use Whoops\Util\Misc;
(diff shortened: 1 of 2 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -532,4 +538,14 @@ public function setResourcesPath($resourcesPath)
...
         $this->addResourcePath($resourcesPath);
     }
+
+    /**
+     * Caster that removes all internals.
+     *
+     * @see AbstactCloner
+     * @return array
+     */
+    public function castObject($obj, $a, $stub, $isNested) {
```
staabm: Could we use a closure instead so we dont need this additional public api method?

staabm:
Instead of only showingnthe most top lvl could we get a bit more detailed and inspect at least 1 level deeper?

jonasdt:
That was also my first idea last week, but it's not so easy ...

src/Whoops/Util/TemplateHelper.php
```diff
@@ -110,10 +116,9 @@ public function dump($value)
         $dumper = $this->getDumper();
 
         if ($dumper) {
-            $cloner = new VarCloner();
 
             // re-use the same DumpOutput instance, so it won't re-render the global styles/scripts on each dump.
-            $dumper->dump($cloner->cloneVar($value), $this->htmlDumperOutput);
+            $dumper->dump($this->cloner->cloneVar($value), $this->htmlDumperOutput);
```
denis-sokolov: Perhaps this is supposed to be `this->getCloner()`?

jonasdt: Correct, I've fixed it.

denis-sokolov:
Did you close it by accident?

jonasdt:
Nope, I'll make a new pull request.

PR #404, opened 2016-03-20
Don't dump object internals

jonasdt:
This improves #402 as follows:

- Only dump object internals if a custom caster exists. This allows integrations with frameworks to define custom casters. E.g. for Laravel we can dump the attributes of an Eloquent model.
- Don't dump verbose information (e.g. exception stack traces).


rossbearman:
This change makes a dramatic speed improvement, from around one minute to under a second, assuming this doesn't cause any issues it would be great to see this pulled into the core.

denis-sokolov:
Thank you, @jonasdt!

PR #415, opened 2016-05-06
Make code-block use tab size of 4 instead of browser default of 8

websiteduck:
Many browsers use a default tab width of 8.  This change will reduce it to 4, which I think is more common for code indentation.


denis-sokolov:

PR #420, opened 2016-06-02
Fixes 419

prisis:


staabm:
in case we start to support symfony/var-dumper < 3.x this should be reflected in the composer.json

prisis:
What is the best way to do it?

Should i do something like ^2.6 || ^3.0 ?

staabm:
I guess so.

prisis:
Added a new matrix for php 5.6 for lowest install

.travis.yml
```diff
@@ -9,12 +9,22 @@ php:
...
+      env: SETUP=lowest
+
 cache:
   directories:
     - "$HOME/.composer/cache"
 
 before_script:
-  - composer install --no-interaction --prefer-dist
+  - if [[ "$SETUP" = "basic" ]]; then composer install --no-interaction --prefer-dist; fi
+  - if [[ "$SETUP" = "lowest" ]]; then travis_retry composer update --prefer-dist --no-interaction --prefer-lowest --prefer-stable; fi
```
staabm: why the travis_retry in this particular case?

src/Whoops/Util/TemplateHelper.php
```diff
@@ -119,8 +119,15 @@ public function dump($value)
         if ($dumper) {
             // re-use the same DumpOutput instance, so it won't re-render the global styles/scripts on each dump.
             // exclude verbose information (e.g. exception stack traces)
+            if (class_exists('Symfony\Component\VarDumper\Caster\Caster')) {
+                $cloner = $this->getCloner()->cloneVar($value, Caster::EXCLUDE_VERBOSE);
```
staabm: this var should be renamed to `clone`

staabm:
thanks looks mostly good!

src/Whoops/Util/TemplateHelper.php
prisis: done :)

prisis:
@filp any info for this pr?

prisis:
@filp any news for this?

denis-sokolov:
Thank you for your contribution, this seems good.

PR #422, opened 2016-06-30
Fix issues with frameworks like Laravel on PHP7

xwiz:
Allow exception or throwable to be passed to whoops


src/Whoops/Exception/Inspector.php
```diff
@@ -58,7 +58,7 @@ public function getExceptionMessage()
     }
 
     /**
-     * Does the wrapped Exception has a previous Exception?
+     * Does the wrapped mixed have a previous Exception?
```
hultberg: Hum, is this an intended change?

src/Whoops/Exception/Inspector.php
```diff
@@ -6,12 +6,12 @@
 
 namespace Whoops\Exception;
 
-use Exception;
+use Throwable;
```
prisis: Wrong | no need for this

src/Whoops/Exception/Inspector.php
```diff
@@ -6,12 +6,12 @@
...
 
 class Inspector
 {
     /**
-     * @var Exception
+     * @var Throwable
```
prisis: Use \Exception | \Throwable

src/Whoops/Exception/Inspector.php
```diff
@@ -26,15 +26,15 @@ class Inspector
     private $previousExceptionInspector;
 
     /**
-     * @param Exception $exception The exception to inspect
+     * @param mixed $exception The exception to inspect
```
prisis: Wrong use \Exception | \Throwable

prisis:
Hey @xwiz you pr is wrong you need to remove all mixed and change it to \Exception | \Throwable and only remove Exception in functions.

src/Whoops/Exception/Inspector.php
xwiz: Ah, thanks for the catch.. Pushed changes now...

prisis:
But wait a min, why you dont use the v2?

GrahamCampbell:
This is a major breaking change.

GrahamCampbell:
That's why it only happened on v2.

GrahamCampbell:
Upgrade to newer laravel to use php 7. Laravel 5.1 or upwards is needed.

GrahamCampbell:
> Fix issues with frameworks like Laravel on PHP7

Laravel 4 does NOT support PHP 7.

xwiz:
@prisis v1 is included by Laravel 4.2 by default. Technically, I don't see anything wrong with Laravel 4.2 and many have gone as far as optimizing it by including their own packages and stuff. I don't see why anyone who decides or wants to continue using it can't use it.

xwiz:
@GrahamCampbell how do you mean it's a major breaking change? What kind of test would fail?

GrahamCampbell:
Anyone extending this class will break.

GrahamCampbell:
> Technically, I don't see anything wrong with Laravel 4.2 

I can guarantee you L4.2 does not work properly on PHP 7. I made the needed changes to Laravel my self to get L5.1 to work.

xwiz:
@GrahamCampbell Interesting. I've heard that it's not compatible, but I don't exactly know what's not compatible except for the error handling which is pretty easy to resolve if you edit the files directly.

GrahamCampbell:
There are about 40 places that need to change to get it to work.

xwiz:
Wow hmmm, thanks for the information. Do you have a fork I can look at? Would be very helpful

GrahamCampbell:
Sorry, no. I'd recommend upgrading to L5.1 LTS.

xwiz:
@GrahamCampbell Me as well. I just wanted to see out of curiosity. Also I prefer 5.2

xwiz:
One issue I don't like with Laravel 5 is the injected/default Validator. I think that belongs to the base Model class, and Laravel should also support defining rules via arrays at the base model class. This is me just saying, not like I've taken it up. I believe every developer should do whatever they want with a framework but there're ways that can be argued to be better in achieving an aim.

prisis:
@xwiz can you close this pr?

GrahamCampbell:
> Also I prefer 5.2

I meant upgrade to at least 5.1.

PR #431, opened 2016-07-28
Remove final flag on the Run class

inssein:
I wanted to extend whoops, and got caught in several places:

1) I implemented the whoops interface, but couldn't use the class as the Handler's expect an implementation of `Run`, not `RunInterface`
2) The run class was final

If there is good reason for this class to be final, the project should atleast type to the interface.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -15,7 +15,7 @@ use Whoops\Handler\HandlerInterface;
 use Whoops\Util\Misc;
 use Whoops\Util\SystemFacade;
 
-final class Run implements RunInterface
+class Run implements RunInterface
 {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Wasn't this already discussed, and final was done on purpose to prevent poor architecture.

inssein:
Makes sense, but if thats the case, the HandlerInterface should accept a RunInterface, not a Run implementation.

denis-sokolov:
Thank you for your pull request.  This will not be accepted, however, see #364.

Your note about the `HandlerInterface` is correct, it should accept a `RunInterface` and not `Run`. A PR doing that will be gladly accepted!

Issue #430, opened 2016-07-27
Monolog & Whoops - Logging errors we don't want to display


inssein:
I am currently working on a fairly old codebase and trying to improve it, but I am stuck on figuring out how to make Whoops work with Monolog.

I integrated Whoops a few months ago, and it has been working out fairly well (for dev and prod). But now, I need to integrate Monolog, and I am having a lot of difficulties as I can't find a good way to do this.

From Googling, everyone who wants to integrate whoops with monolog adds monolog as a handler to whoops.  However, I have a few issues with that.

1. I want to log all errors, but only display fatal errors / unhandled exceptions in production
2. The monolog ErrorHandler seems to add better information about the calling function and its parameters.

So what I did was I registered both separately. This is almost there, except I've got another two issues:

1. Whoops is displaying every error that is reported. How can I limit this without changing error_reporting as I want to log all errors?
2.  After monolog logs a PHP error, Whoops handles the same error, converts it to an exception, and then Monolog again catches it.. (causing double logging).

I can't think of a clean way to solve the above issues. Is the solution just to write a separate production error handler and not use whoops in prod (if I want to log everything but not display everything)?

Edit:
The solution might be as simple as disabling error handling in prod.

inssein:
Another question:

Why does Whoops throw the exception, instead of just handling it?

As in, why do this:
```PHP
$exception = new ErrorException($message, /*code*/ $level, /*severity*/ $level, $file, $line);
if ($this->canThrowExceptions) {
    throw $exception;
} else {
    $this->handleException($exception);
}
```
Instead of:
```
$exception = new ErrorException($message, /*code*/ $level, /*severity*/ $level, $file, $line);
$this->handleException($exception);
```

denis-sokolov:
We throw exceptions for consistency, to handle everything in the same way. This also allows catching those exceptions in user code.

I don't know how to help you with the main issue. If you have more concrete ideas how to improve Whoops to allow for your case, let us know. A PR would be even better. Until then, this is hardly actionable. :(

PR #432, opened 2016-07-29
Fix extensibility

inssein:
Since the `\Whoops\Run` class is final, extending it is a no go, but we should then be able to write our own implementation of it. However, the handlers expect an `\Whoops\Run` instead of `\Whoops\RunInterface`, which stops me from extending Whoops completely.

All that said and done, The RunInterface itself is pretty bloated. Things like `silenceErrorsInPaths` should probably be details of a given implementation, and not in the interface. (For example, my own implementation wouldn't care about that functionality, but I have to implement that function due to the interface contract). I haven't addressed this in this PR, but I could if it we agree.

Edit: This is part of https://github.com/filp/whoops/issues/430

### DIFF
--- a/src/Whoops/Handler/Handler.php
+++ b/src/Whoops/Handler/Handler.php
@@ -7,7 +7,7 @@
 namespace Whoops\Handler;
 
 use Whoops\Exception\Inspector;
-use Whoops\Run;
+use Whoops\RunInterface;
 
 /**
  * Abstract implementation of a Handler.
@@ -24,7 +24,7 @@ abstract class Handler implements HandlerInterface
     const QUIT         = 0x30;
 
     /**
-     * @var Run
+     * @var RunInterface
      */
     private $run;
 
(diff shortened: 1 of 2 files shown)


denis-sokolov:
Thank you for your contribution! I fully agree that the `RunInterface` is bloated. A PR on that would be welcome. Until then, we, unfortunately, need to have dummy implementations of these methods.

staabm:
Changing it would require a new major version

denis-sokolov:
Why? Relaxing an interface without changing our `Run` and our `Handler`s should not break anything for others. What would be a scenario where that would break?

staabm:
Client code could rely on methods beeing present and call those (which works before you relax the interface and breaks afterwards)

denis-sokolov:
But only if both client code AND `Run` Imlpementation are customized, and the methods are removed after the change. Why would someone explicitly do that for their own code? Our `Run` will keep those methods, even if they are not present in the interface.

inssein:
I can see how it would require a major version as it could be a pretty big BC break, but considering you couldn't have your own Run implementation until this PR, I don't know if it is.

Issue #435, opened 2016-08-06
PrettyPageHandler - code highlight breaking


acidjazz:
It's breaking around comments

![image](https://cloud.githubusercontent.com/assets/967369/17459425/dd9bd310-5beb-11e6-986d-9e2c4993507b.png)

prisis:
Hey :), duplicate of https://github.com/filp/whoops/issues/426, take the search first, next time

acidjazz:
@prisis I did, and saw those, but they're all closed, so I assume they're fixed?

prisis:
@acidjazz if you take a look in https://github.com/filp/whoops/issues/214, we didn't find a good solution to fix this...

acidjazz:
I see @prisis , interesting, what is being used to highlight php code?

prisis:
whoops use https://github.com/filp/whoops/blob/a9d17d13aca43d9b39aa5073d64ff4086b955946/src/Whoops/Resources/js/whoops.base.js#L15

acidjazz:
@Prisis looks like https://github.com/google/code-prettify

acidjazz:
I've written a fix for this:

![image](https://cloud.githubusercontent.com/assets/967369/17460003/2741eafc-5c07-11e6-9f0b-ad9ecd2a646a.png)

What it does is allow the prettifier to highlight the entire file, then i set a maximum height on the div and scroll to the line of code, this allows also the functionality to browser the entire file.

Let me know what you think I'll create a pull request.

acidjazz:
@prisis 
https://github.com/filp/whoops/pull/436

filips123:
@acidjazz @denis-sokolov This still sometimes not work. See #588 and google/code-prettify#553 for details.

Issue #426, opened 2016-07-23
Inverted syntax coloration for comments and code


Litarvan:
The code syntax coloration is inverted. Code is considered as comment, and comment as code. Only when there is doc comment before. It does not happen everytime but often.

![Screen](https://i.gyazo.com/7f5830a871cc95faf7578c485548239b.png)

Some informations : 

SERVER_SOFTWARE "PHP 5.6.8 Development Server"
SERVER_PROTOCOL "HTTP/1.1"
SERVER_NAME "localhost"
SERVER_PORT "58777"
REQUEST_URI "/test"
REQUEST_METHOD "GET"
SCRIPT_NAME "/index.php
HTTP_HOST "localhost:58777"
HTTP_USER_AGENT "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0"

denis-sokolov:
Thank you for your report. We have already discussed this in #214, and, since there is no easy obvious fix, and the impact of the bug is small, we have decided to not do anything about it. If anyone has motivation to fix this, a PR is most welcome. Otherwise, this will be a small quirk in Whoops for a while. :)

Litarvan:
Ok thanks, maybe I will send a post request ^^

Issue #385, opened 2016-02-15
Syntax highlighting wrong when first line of snippet is the end of a block comment


CodeBrauer:
![bildschirmfoto 2016-02-15 um 17 44 31](https://cloud.githubusercontent.com/assets/2059754/13054876/4af53c84-d40c-11e5-974e-d36c7f7ce947.png)

Line 22 is the end of a block comment and breaks the highlighting

staabm:
duplicate of https://github.com/filp/whoops/issues/214

denis-sokolov:
Thanks for the report! Thanks for the link, @staabm!

filips123:
@acidjazz @denis-sokolov This still sometimes not work. See #588 and google/code-prettify#553 for details.

Issue #214, opened 2014-05-27
Syntax highlighting breaks if the code starts with a partial comment


denis-sokolov:
![1](https://cloud.githubusercontent.com/assets/113721/3089841/f9030924-e583-11e3-892f-14160bc1fee0.png)

We should be able to detect such situations and workaround them.

GRMule:
I had an idea that I thought might work. After `$range = $frame->getFileLines($line - 8, 10);`, check the first line, and if the first non-space char is an asterisk, then get a new range including one more line above and check again. I figured you could do this a set number of times, then give up so as to not make the code sample too long in the name of chasing the comment head. Or, as ConnorVG pointed out, if the error IS the comment head then you'll never finish, so there would have to be a threshold for giving up.

This is my initial attempt at that idea:

````php
$firstLine = trim(reset($range));
$firstCharIsAsterisk = $firstLine[0] === '*';
$giveup = 10; // give up after 10 tries to find the comment head
$tryCtr = 0;
while ($firstCharIsAsterisk !== false && $tryCtr < $giveup) {
    $range = $frame->getFileLines($line - 8 - $tryCtr, 10 + $tryCtr);
	$firstLine = trim(reset($range));
	$firstCharIsAsterisk = $firstLine[0] === '*';
	$tryCtr++;
}
````
The problem is that intermediary lines of multiline comments don't *have* to start with an asterisk. My idea worked with this comment:

````php
			/*
			 * comment 1
			 * comment 2
			 * comment 3
			 * comment 4
			 * comment 5
			 * comment 6
			 * comment 7
			 */
			$t = 1/0; // exception here
````

I got a larger code sample window that included the error line all the way up to the comment head. You could also just use that code to verify that there IS a comment head somewhere in there, then go with the original range but append the opening comment as filp mentioned. 

This comment, however, broke my idea:
````php
			/*
			  comment 1
			  comment 2
			  comment 3
			  comment 4
			  comment 5
			  comment 6
			 */ 
			$t = 1/0; // exception here
````

Most people use the asterisk all the way down, but it isn't necessary. Without that as a hint, there's really no way I can think of to evaluate whether a given line, in isolation, is a comment or not. As far as I can see right now, you'd have to tokenize the file and go from the start down to where the error occurred and examine the tokens on the way... yikes!

Thought I would share my efforts in case this helps someone else think of a solution, or at least see what does *not* work.

denis-sokolov:
[...]

PR #436, opened 2016-08-07
Fixes the syntax highlighting issue

acidjazz:
This addresses #435, #214, #426, and #385 

From
image

To

image

What it does is allow the prettifier to highlight the entire file, then i set a maximum height on the containing div and scroll to the line of code, this could also allow the functionality to browse the entire file if we wanted, but would show scrollbars on windows machines.

### DIFF
--- a/src/Whoops/Resources/js/whoops.base.js
+++ b/src/Whoops/Resources/js/whoops.base.js
@@ -65,9 +65,9 @@ Zepto(function($) {
 
   });
 
+  // Scroll to the active
   function scrollToLine() {
 
-    var $activeLine     = $frameContainer.find('.frame.active');
     var activeLineNumber = +($activeLine.find('.frame-line').text());
     var id     = /frame\-line\-([\d]*)/.exec($activeLine.attr('id'))[1];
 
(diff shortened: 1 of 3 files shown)


src/Whoops/Resources/js/whoops.base.js
```diff
@@ -1,4 +1,6 @@
 Zepto(function($) {
+
```
GrahamCampbell: why the extra newlines?

denis-sokolov:
This seems to be a great job, thanks for your contribution!

What's the performance of that on huge files, like 2000-3000 lines of code? While I think such files are terribly too long, in real life people do have files like that.
Could you check whether a long file causes any visible hiccup?

acidjazz:
@denis-sokolov excellent question, I did look into the prettifier and it does have a limit itself: https://github.com/google/code-prettify/issues/44

denis-sokolov:
158k? O_o Alright, that's irrelevant for all practical purposes. :)
Thank you!

filips123:
@acidjazz This still sometimes not work. See #588 and google/code-prettify#553 for details.

Issue #438, opened 2016-08-10
PrettyPageHandler - optimize googles prettify to only render viewable area


acidjazz:
This is an optimization I found while researching the functionality of googles prettify.js, pull request incoming

acidjazz:
addressed/fixed in #439

Issue #437, opened 2016-08-10
PrettyPageHandler - resizing the window horizontally alters the <pre> scroll offset


acidjazz:
Discovered after introducing pull request #436 . Fix incoming via pull request

acidjazz:
fixed in pull request #439

PR #439, opened 2016-08-10
bug fixes and prettyprint optimization

acidjazz:
addresses #438, #437

I did some more testing of my fix in pull #436 after @denis-sokolov suggested potential performance issues with prettifying larger source files, I found that files of 1k+ lines of code started to take a toll. I also did some playing around with googles [code-prettify](https://github.com/google/code-prettify) and added some optimizations.

* Instead of rendering the entire file in each code block, I only render 20 lines above and below, this can be increased if we start seeing larger multi-line comment blocks, I can even make it a configurable option if needed.

* I noticed that prettyPrint() will render every element having the class `prettyprint`, which was also prettifying all the hidden code-blocks in the entire stack, so I altered this to only render the visible blocks of code
 * Upon clicking a stack frame, the `prettyprint` class is swapped from the old to the new
 *  prettyPrint() is then run again
 * I found prettyPrint() also has a callback for when its complete, and moved my functionality into that instead of a timeout, speeding up rendering a little bit more

This decreased my render time drastically, especially on errors displayed with larger stacks.

Other Details
 * Exchanged the css properties applied to the `prettyprint` classes with `code-block`, this way swapping this class did not alter the current look and feel
 * Removed excess inline css in the code-block view that was manually highlighting, this is already done in highlightCurrentLine()
 * Updated the prettyprint cdn url to the latest version
 * Adjusted my function structure to match the existing

### DIFF
--- a/src/Whoops/Handler/JsonResponseHandler.php
+++ b/src/Whoops/Handler/JsonResponseHandler.php
@@ -40,10 +40,12 @@ class JsonResponseHandler extends Handler
     public function handle()
     {
         $response = [
-            'error' => Formatter::formatExceptionAsDataArray(
+          'errors' => [
+            Formatter::formatExceptionAsDataArray(
                 $this->getInspector(),
                 $this->addTraceToOutput()
             ),
+          ],
         ];
 
         if (\Whoops\Util\Misc::canSendHeaders()) {
--- a/src/Whoops/Handler/JsonResponseHandler.php
+++ b/src/Whoops/Handler/JsonResponseHandler.php
@@ -40,12 +40,10 @@ class JsonResponseHandler extends Handler
     public function handle()
     {
         $response = [
-          'errors' => [
-            Formatter::formatExceptionAsDataArray(
+            'error' => Formatter::formatExceptionAsDataArray(
                 $this->getInspector(),
                 $this->addTraceToOutput()
             ),
-          ],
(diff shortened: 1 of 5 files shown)


acidjazz:
@denis-sokolov just a quick update, been using this version for a while now w/ all my projects and its definitely a lot faster and more stable than when I was prettifying the entire file.

acidjazz:
@denis-sokolov quiet ping about this PR, still a lot more stable so far on my end

denis-sokolov:
I'm sorry about the delay, I'll be back from my vacation this weekend.

staabm:
You should not mix a PR with patches for different topics.

In general I like the JsonResponse patch, but it needs to be done in a BC manner, e.g. The new default cant either be the default or it needs to be a new Handler

acidjazz:
@staabm sure,  what do you mean by BC? And which direction do you suggest?

staabm:
@acidjazz BC means backwards compatibility. we cannot change the default format because this will break for everyone which relies on it right now.

ATM I would add a new method on the JsonResponseHandler which effectively activates the "JsonAPI spec compilance mode".

acidjazz:
@staabm ok, reverting all my changes for the feature, and adding a new method

denis-sokolov:
Thank you, this works!

PR #445, opened 2016-09-04
Plain text in pretty page

### LINKED ISSUE #434: Suggestion, add stacktrace in a HTML comment on top of document (PrettyPageHandler)
Sometimes I get the prettty page back while I can only view the HTML source itself. I always find myself scrolling down through all the CSS code to find the stacktrace. Maybe it's an idea to add the stacktrace to a HTML comment on top of the prettypage output.

denis-sokolov:
Fixes #434.


staabm:
Did not test it right now, but looks good

PR #446, opened 2016-09-05
Fix private properties

wawan93:
Replace all private properties to protected, to allow override handlers and exceptions


staabm:
Whats your use case? Just changing all the properties isnt a good idea.

GrahamCampbell:
👎  on this.

wawan93:
@staabm I need to extend PlainTextHandler and JsonResponseHandler in my project, but I can't because I can't override private methods and properties.

staabm:
@wawan93 could you describe in more detail what are you trying to achieve?

denis-sokolov:
The variables are private on purpose, we do not support extending classes, see #364 and feel free to provide new information if you have any.

PR #449, opened 2016-09-10
Add IntelliJ IDEA support

msheakoski:
Some of us use IntelliJ IDEA with the PHP plugin instead of PhpStorm.

I'm aware that setEditor() takes a callable, but using a known editor makes it much easier to configure. A twelve-factor app can simply check the environment for `WHOOPS_EDITOR=idea` and pass that value directly into PrettyPageHandler instead of adding a bunch of logic to convert it into a callable.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -76,6 +76,7 @@ class PrettyPageHandler extends Handler
         "emacs"    => "emacs://open?url=file://%file&line=%line",
         "macvim"   => "mvim://open/?url=file://%file&line=%line",
         "phpstorm" => "phpstorm://open?file=%file&line=%line",
+        "idea"     => "idea://open?file=%file&line=%line",
     ];
 
     /**


msheakoski:
I guess the logic isn't as bad as I thought for supporting built-in + custom editors:
```php
$whoops = new Run();
$prettyPageHandler = new PrettyPageHandler();
$whoopsEditor = getenv("WHOOPS_EDITOR") ?: "";
if ($whoopsEditor !== "") {
    if (strpos($whoopsEditor, "://") !== false) {
        $prettyPageHandler->setEditor(function () use ($whoopsEditor) {
            return $whoopsEditor;
        });
    } else {
        $prettyPageHandler->setEditor($whoopsEditor);
    }
}
$whoops->pushHandler($prettyPageHandler);
```
vs only supporting built-in editors:
```php
$whoops = new Run();
$prettyPageHandler = new PrettyPageHandler();
if ((getenv("WHOOPS_EDITOR") ?: "") !== "") {
    $prettyPageHandler->setEditor(getenv("WHOOPS_EDITOR"));
}
$whoops->pushHandler($prettyPageHandler);
```
It would still be nice to have IntelliJ IDEA built in so you can simply say "idea" vs the entire URI.

PR #450, opened 2016-09-18
Whoops for PHP 5.3.3 infinite loop caused by zlib.compression_output

kaizendeveloper:
On PHP 5.3.3 (FPM) if zlib.output_compression is "On"  doing ob_end_clean() inside of Run.php will launch an error (failed to delete buffer zlib output_compression) that will be caught by the very handler at the same time, looping forever inside of the while block, if the user can't disable zlib.output_compression by any means or if is it needed by any other reason a workaround that I used myself to get whoops functioning with this PHP version is not to call ob_end_clean(), but exiting the while block and let the flush() instruction below do its work.


denis-sokolov:
There is something wrong with this PR. It seems there has been a mistake, as there is no reason to merge `master` into `v1`. Feel free to create a new PR if you can clarify your intent.

PR #451, opened 2016-09-20
Add silencedPatterns getter

funivan:
For example if we want to change logic of the `Whoops\Run::handleError` we can extend `SystemFacade` and rewrite method `setErrorHandler`. But unfortunately we can`t filter errors inside our custom error handler.


denis-sokolov:

PR #456, opened 2016-10-04
Update composer.json

vinkla:
1. Removed default homepage url (packagist will use the github url by default)
2. Update the php version requirement

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,7 @@
     "license": "MIT",
     "description": "php error handling for cool kids",
     "keywords": ["library", "error", "handling", "exception", "whoops", "zf2"],
+    "homepage": "https://filp.github.io/whoops/",
     "authors": [
         {
             "name": "Filipe Dobreira",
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Is the intent with the version change to not mistakenly claim we support PHP 8?

composer.json
```diff
@@ -3,7 +3,6 @@
     "license": "MIT",
     "description": "php error handling for cool kids",
     "keywords": ["library", "error", "handling", "exception", "whoops", "zf2"],
-    "homepage": "https://github.com/filp/whoops",
```
staabm: we should instead use https://filp.github.io/whoops/ as the url

vinkla: Oh, yeah. True! I'll fix this.

vinkla:
@denis-sokolov yeah, we don't know if the library will work on PHP 8.0 once it is released. This way we can control that.

PR #457, opened 2016-10-10
Deleted status code verification

izabolotnev:
Why I can't use any status code? I.e. I want wrap Json-RPC error object and return it with status code 200 but I can't.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -203,12 +203,6 @@ final class Run implements RunInterface
             $code = 500;
         }
 
-        if ($code < 400 || 600 <= $code) {
-            throw new InvalidArgumentException(
-                 "Invalid status code '$code', must be 4xx or 5xx"
-            );
-        }
-
         return $this->sendHttpCode = $code;
     }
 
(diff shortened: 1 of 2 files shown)


staabm:
does your framework require a error object beeing transmitted with a http 200?
(doesnt make much sense IMO)

denis-sokolov:
If you need a workaround for whatever purpose, you are welcome to disable status code sending and send it yourself `$run->sendHttpCode(false);`.

But for canonical use cases, the HTTP status code in case of a server error should be in 500s, as per HTTP spec.

denis-sokolov:
That said, thank you for your contribution and an attempt on making Whoops better! The fact I disagree with this PR does not subtract from the quality of the contribution! :)

zinovyev:
A pity that you've rejected this PR. We really need this to communicate via JsonRPC 2.0 properly (the response would be with 200 code, but the message will contain a real error code).
Thank you for your workaround anyway.

denis-sokolov:
You seem to know what you’re doing, and you will have a workaround that works. In the meantime, most users using Whoops should be guided into a more correct route.

Issue #469, opened 2016-12-25
Broken header sending


GrahamCampbell:
The pretty page handler is screwing with the headers, even when `writeToOutput(false);` is called. This is because of https://github.com/filp/whoops/commit/d7d89b0cce35b7dff10aae14d5b075c3c6acd610. That commit should be reverted because it doesn't work, and screws with the global state without being asked.

staabm:
There are more handlers then the prettyPage handler which do this very same step. Should we check the writeToOutput() flag in those sports?

GrahamCampbell:
I have a fix coming very soon...

GrahamCampbell:
There are some other issues that will become apparent with my PR too, when we have more than one handler, we get two of the same header, or more, sent, which screws up some servers, as well as being incorrect.

GrahamCampbell:
@staabm Here we go: https://github.com/filp/whoops/pull/470.

staabm:
> when we have more than one handler, we get two of the same header, or more, sent, which screws up some servers, as well as being incorrect.

Could you elaborate? Can several handlers be invoked for rendering a response for 1 request?

GrahamCampbell:
> Could you elaborate? Can several handlers be invoked for rendering a response for 1 request?

Yes, the while loop continues until a handler says "quit".

GrahamCampbell:
Meh, not a while loop, for loop.

staabm:
Yeah, but the new impl in your PR will send the headers only once, right? So this problem is gone?

GrahamCampbell:
Yes. My PR fixes all the problems I can see with the headers.

denis-sokolov:
Closed in #470.

PR #470, opened 2016-12-25
Deal with header sending correctly

GrahamCampbell:
See https://github.com/filp/whoops/issues/469 and https://github.com/GrahamCampbell/Laravel-Exceptions/pull/92.


staabm:
Should have a test

staabm:
Looks good to me.

PR #474, opened 2017-01-22
Improved PrettyPageHandler::getEditor

sergeyklay:
Reduced cyclomatic complexity for `PrettyPageHandler::getEditor`. Now `PrettyPageHandler::getEditor` always return array

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -390,7 +390,7 @@ class PrettyPageHandler extends Handler
     {
         $editor = $this->getEditor($filePath, $line);
 
-        if (!$editor) {
+        if (empty($editor)) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -436,42 +436,37 @@ public function getEditorAjax($filePath, $line)
...
+            return [];
         }
-        else if(is_string($this->editor) && isset($this->editors[$this->editor]) && !is_callable($this->editors[$this->editor]))
-        {
+
+        if (is_string($this->editor) && isset($this->editors[$this->editor]) && !is_callable($this->editors[$this->editor])) {
            return [
                 'ajax' => false,
-                'url' => $this->editors[$this->editor],
+                'url'  => $this->editors[$this->editor],
```
denis-sokolov: Usually you don't want to add styling changes to other peoples' code. In some cases it might be difficult to distinguish a style difference from a clear mistake. In this case, the decision to align the values in the array, is purely a stylistic choice. In Whoops we decided not to. Personally, I believe our way is _clearly, objectively_ better, but that's the tricky part with style. :)

But, even more importantly, you don't want to submit these style changes together with other changes in a PR. The developers might want to accept your other changes, but not like the style changes. And that is why every change belongs in its own PR - to allow discussing it individually.

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -436,42 +436,37 @@ public function getEditorAjax($filePath, $line)
...
-            else
-            {
+            } else {
                 $callback = call_user_func($this->editors[$this->editor], $filePath, $line);
             }
 
             return [
                 'ajax' => isset($callback['ajax']) ? $callback['ajax'] : false,
-                'url' => (is_array($callback) ? $callback['url'] : $callback),
+                'url'  => isset($callback['url']) ? $callback['url'] : $callback,
```
denis-sokolov: `isset($callback['url'])` apparently works and correctly returns `false` if `$callback` is a string. But it's real unexpected for me to see `isset` called on a string.

Would you please clarify this by adding the following statement on line 462:

```php
if (is_string($callback)) {
    $callback = [
        'ajax' => false,
        'url' => $callback;
    ];
}
```

sergeyklay: As I can see just return enough :)

```php
if (is_string($callback)) {
    return [
        'ajax' => false,
        'url'  => $callback,
    ];
}
```

src/Whoops/Handler/PrettyPageHandler.php
sergeyklay: Totally agree man :)

sergeyklay:
@denis-sokolov Could you please take a look

staabm:
Hmm should we handle this like a BC breaking change?

denis-sokolov:
@staabm, can you clarify what’s the breaking scenario?

staabm:
Just realized that this PR does not change signature but only fixes wrong phpdocs :-). 

Everything fine then.

denis-sokolov:
@staabm, thank you for being vigilant.

@sergeyklay, I will keep this open for a day or two to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me!

PR #475, opened 2017-01-23
Simplified SystemFacade

sergeyklay:
* Improved `SystemFacade::setErrorHandler`
* Fixed return type for few methods

### DIFF
--- a/src/Whoops/Util/SystemFacade.php
+++ b/src/Whoops/Util/SystemFacade.php
@@ -20,23 +20,19 @@ class SystemFacade
 
     /**
      * @param callable $handler
-     * @param int      $types
+     * @param int      $types By default E_ALL | E_STRICT
      *
-     * @return callable|null
+     * @return string|null
      */
-    public function setErrorHandler(callable $handler, $types = 'use-php-defaults')
+    public function setErrorHandler(callable $handler, $types = null)
     {
-        // Workaround for PHP 5.5
-        if ($types === 'use-php-defaults') {
-            $types = E_ALL | E_STRICT;
-        }
-        return set_error_handler($handler, $types);
+        return set_error_handler($handler, $types === null ? E_ALL | E_STRICT : $types);
     }
 
     /**
      * @param callable $handler
      *
-     * @return callable|null
+     * @return string|null
      */
     public function setExceptionHandler(callable $handler)
     {
--- a/tests/Whoops/Util/SystemFacadeTest.php
+++ b/tests/Whoops/Util/SystemFacadeTest.php
@@ -156,12 +156,8 @@ function flush()
     return SystemFacadeTest::delegate('flush');
 }
 
-function set_error_handler(callable $handler, $types = 'use-php-defaults')
+function set_error_handler(callable $handler, $types = null)
 {
-    // Workaround for PHP 5.5
-    if ($types === 'use-php-defaults') {
-        $types = E_ALL | E_STRICT;
-    }
     return SystemFacadeTest::delegate('set_error_handler', func_get_args());
 }
 


denis-sokolov:
What is the added benefit for all new returned values?

sergeyklay:
In general, I think it's a good idea when a method always returns something. For example see https://github.com/sergeyklay/whoops/blob/b3a858e5040e0439faad46acb34ce0f0edc61abc/tests/Whoops/Util/SystemFacadeTest.php#L134-L197 we always expect something, isn't it? I believe that there is nothing wrong or unnecessary to always return a result. Moreover, the original function will do exactly that.

As I see the tests fails. I'll take a look later

denis-sokolov:
I disagree that it’s a good idea to always return something. If we have nothing meaningful to tell, better tell nothing, to avoid creating false meaning where there is none.

For an example, if I call a `registerShutdownFunction` that returns `void`, it’s clear: I pass it a callback and I’m done. There is nothing else I can do. But if the same function now returns `mixed`, not only I need to find out what that `mixed` is, but also what am I supposed to do with it. Do I need to handle an error of some sort? Do I need to store a reference or index I get back? These are questions that I don’t want to care about.

The code example in `SystemFacadeTest` that you have linked to is a red herring. All of those functions are delegates or proxies. Their entire point is they don’t care about the return value or the absence of one, they simply forward it along.

Functions in PHP standard library do all kinds of stuff. It’s hardly a stellar example of consistent API design. As valuable sticking to existing conventions is, there are no conventions in PHP standard library. :)

staabm:
Agree, that we either need a proper use-case for the introduced return-values, or should revert that part. 

Our objects should hide their internal  state instead

prisis:
Maybe we should add @internal on some functions

And I don't like the new return types too.

sergeyklay:
So.
After some thought I came to the conclusion that I completely agree with @denis-sokolov. I removed all unneeded returns and rolled back  all returned values.

@denis-sokolov could you take a look now?

src/Whoops/Util/SystemFacade.php
```diff
@@ -20,23 +20,19 @@ public function startOutputBuffering()
...
-     * @param int      $types
+     * @param int      $types By default E_ALL | E_STRICT
      *
-     * @return callable|null
+     * @return string|null
      */
-    public function setErrorHandler(callable $handler, $types = 'use-php-defaults')
+    public function setErrorHandler(callable $handler, $types = null)
     {
-        // Workaround for PHP 5.5
```
denis-sokolov: If I blame the removed code, I find that this workaround was put here on a very explicit purpose in e18d217. Whilst the proposed changes do not fail anymore, the note about it being a workaround is gone – it is easy to expect the next developer "simplifying" the code further and reverting it to the pre-workaround version.

Instead, I would rather keep the very explicit conditional.

src/Whoops/Util/SystemFacade.php
```diff
@@ -20,23 +20,19 @@ public function startOutputBuffering()
...
-        }
-        return set_error_handler($handler, $types);
+        return set_error_handler($handler, $types === null ? E_ALL | E_STRICT : $types);
     }
 
     /**
      * @param callable $handler
      *
-     * @return callable|null
+     * @return string|null
```
denis-sokolov: This is not true. The function implementation returns the return value of PHP built-in `set_exception_handler`. [The documentation](http://php.net/manual/en/function.set-exception-handler.php) states it returns a callable. Something is wrong here, and the proposed change needs much more explanation to be included.

denis-sokolov:
Even though we did not merge this in, I thank you for creating suggestions and discussing them. Integrating software with other people is always work, it’s debate and discussions, so please don’t be discouraged to try other ideas later on!

PR #477, opened 2017-02-04
Whoos v1 is not php7 compatible

staabm:
In v1 php7 doesnt really work because of used Exception type (which doesnt work with Error types)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0"
+        "php": "^5.3.0"
     },
     "require-dev": {
         "mockery/mockery": "0.9.*"


GrahamCampbell:
This change is pretty pointless because of the way dependency resolution works. Composer will just resolve the version exactly one before this commit.

GrahamCampbell:
👎

staabm:
Hmm I guess you are right.. we should add a runtime version check instead

staabm:
Alternatively we could remove the typehints instead

denis-sokolov:
I support any of these solutions.

GrahamCampbell:
I personally don't think any changes should be made at all to v1. It's just fine.

Issue #481, opened 2017-02-11
Mask php-auth-user/password


staabm:
Should we per default mask php auth user/password, which are currently display the user/password as plain-text within the server variables

http://php.net/manual/en/features.http-auth.php

denis-sokolov:
It's a minor, nice feature.  If somebody submits a pull request, I will merge it.  If not, no big deal.

PR #482, opened 2017-02-12
Allow sensitive data contained within one of the superglobal arrays b…

staabm:
…eeing masked.

implements the idea discussed in https://github.com/filp/whoops/issues/481

this adds some very basic mechanics and a api to mask sensitive data within one of the superglobal arrays.
by default the php auth based _SERVER vars will be masked.

I did not add any tests, as I didnt see something worth testing.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -60,6 +60,19 @@ class PrettyPageHandler extends Handler
     private $applicationPaths;
 
     /**
+     * @var array[]
+     */
+    private $blacklist = [
+        '_GET' => [],
+        '_POST' => [],
+        '_FILES' => [],
+        '_COOKIE' => [],
+        '_SESSION' => [],
+        '_SERVER' => [],
+        '_ENV' => [],
+    ];
+
+    /**
      * A string identifier for a known IDE/text editor, or a closure
      * that resolves a string that can be used to open a given file
      * in an editor. If the string contains the special substrings
@@ -98,6 +111,9 @@ class PrettyPageHandler extends Handler
 
         // Add the default, local resource search path:
         $this->searchPaths[] = __DIR__ . "/../Resources";
+
+        // blacklist php provided auth based values
+        $this->blacklist('_SERVER', 'PHP_AUTH_PW');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -98,6 +111,10 @@ public function __construct()
 
         // Add the default, local resource search path:
         $this->searchPaths[] = __DIR__ . "/../Resources";
+        
+        // blacklist php provided auth based values
+        $this->blacklist('_SERVER', 'PHP_AUTH_USER');
```
staabm: here you can see the public api a whoops user would call for this feature in action.

staabm: we could also only ship the mechanics without this 2 vars beeing masked by default.

denis-sokolov: In my opinion, user portion is not that sensitive. In addition, because we do not have an API to unblacklist, it is easy to annoy existing users of Whoops. Thus I advice to remove `PHP_AUTH_USER` from the default blacklist, but keep `PHP_AUTH_PW`.

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -214,13 +231,13 @@ public function handle()
...
-                "Session"               => isset($_SESSION) ? $_SESSION :  [],
-                "Server/Request Data"   => $_SERVER,
-                "Environment Variables" => $_ENV,
+                "GET Data"              => $this->maskBlacklisted($_GET, $this->blacklist['_GET']),
+                "POST Data"             => $this->maskBlacklisted($_POST, $this->blacklist['_POST']),
+                "Files"                 => $this->maskBlacklisted($_FILES, $this->blacklist['_FILES']),
+                "Cookies"               => $this->maskBlacklisted($_COOKIE, $this->blacklist['_COOKIE']),
+                "Session"               => isset($_SESSION) ? $this->maskBlacklisted($_SESSION, $this->blacklist['_SESSION']) :  [],
+                "Server/Request Data"   => $this->maskBlacklisted($_SERVER, $this->blacklist['_SERVER']),
+                "Environment Variables" => $this->maskBlacklisted($_ENV, $this->blacklist['_ENV']),
```
denis-sokolov: There’s a bit too much noise for the simple change in functionality. I would like to ask you to improve this a bit.

The simplest change is to not repeat `this->blacklist`, since `maskBlacklisted` can get it itself: `$this->maskBlacklisted($_GET, '_GET')`. Next, we can rename the function to be simpler, since it’s a private function: `$this->masked($_GET, '_GET')`. Finally, we can remove the duplication if we rely on `$GLOBALS`, which is not a big deal here: `$this->masked('_GET')`, and `$GLOBALS[$name]` in the function.

Feel free to do your own changes, as long as the end result is considerably less noisy than what we see now.

staabm:
Great feedback, thx. didn't though about using $GLOBALS which makes it a small patch.

Adjusted and pushed.

garygreen:
It would be nice if you could block the display of all ENV information by wildcard e.g. `blacklist('ENV', '*');`

staabm:
that way all ENV vars would be rendered with "***" ... does this still make sense or wouldnt it be more helpfull to not render the ENV information at all?

garygreen:
That's true, would be better to just have ability to switch off the ENV panel. I personally don't find much use in displaying it and it's just a risk having that information exposed

PR #484, opened 2017-03-02
added psr/log to required dependencies

### LINKED ISSUE #483: Missing Interface Psr\Log\LoggerInterface
In `PlainTextHandler` you expect an instance of `Psr\Log\LoggerInterface` - may be you have a dependency that is missing in composer.json?

I changed some dependencies in my app (removed phpdocumentor) and suddenly get this error. I solved by require psr/log but you should fix it on your site.

staabm:
Requirement inspired by Monolog. Technically we have a `use` Statement which requires this lib. the remaining phpdoc typehints would work without the lib, but still this seems to be the clean way.

Closes #483.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
         }
     ],
     "require": {
-        "php": "^5.5.9 || ^7.0"
+        "php": "^5.5.9 || ^7.0",
+        "psr/log": "^1.0.1"        
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thanks for a quick fix. I suppose adding standard interfaces does not sacrifice our "no dependencies" feature.

staabm:
same opinion... this lib is meant for interop, therefore a good thing to have.

PR #486, opened 2017-03-06
fixed undefined index $GLOBALS  when auto_globals_jit is enabled

staabm:


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -658,16 +658,17 @@ public function blacklist($superGlobalName, $key) {
      * Checks all values identified by the given superGlobalName within GLOBALS.
      * Blacklisted values will be replaced by a equal length string cointaining only '*' characters.
      *
+     * @param $superGlobal array One of the superglobal arrays
      * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
      * @return array $values without sensitive data
      */
-    private function masked($superGlobalName) {
+    private function masked(array $superGlobal, $superGlobalName) {
```
denis-sokolov: You must clearly mention in this function the problem you are fixing, otherwise in a few months we will "improve" this again by returning `$GLOBALS`.

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -658,16 +658,17 @@ public function blacklist($superGlobalName, $key) {
      * Checks all values identified by the given superGlobalName within GLOBALS.
```
denis-sokolov: This line in the documentation is now outdated, with `GLOBALS` gone.

staabm:
I will cleanup this PR after the fix is confirmed.

magnus-eriksson:
I can confirm that this PR fixed the error.

staabm:
thx magnus!

@denis-sokolov  adjusted per comments

denis-sokolov:
Thank you, @staabm!

Issue #509, opened 2017-06-24
PlainTextHandler: Customizable dump function


zonuexe:
The current implementation dumps value using `var_dump()`.
However, I would like to output by any function (eg. [`var_export()`](http://php.net/manual/function.var-export.php), [Symfony VarDumper](https://symfony.com/doc/current/components/var_dumper.html)).

The following code is a proposal by adding `PlainTextHandler::setDumper()`:

```php
$handler = new \Whoops\Handler\PlainTextHandler();
// PHP 7.1+
$handler->setDumper(Closure::fromCallable('var_export'));
// PHP 5, PHP7.0
$handler->setDumper('var_export');

$whoops = new \Whoops\Run;
$whoops->pushHandler($plain_text_handler);
```

staabm:
Btw: the symfony var-dumper is already supported

zonuexe:
@staabm 
Yes.  Symfony\VarDumper has `VarDumper::dump()` method that can be extended by users.

We can not use that method for `PlainTextHandler`.  There is a possibility that [HtmlDumper](https://github.com/symfony/var-dumper/blob/master/Dumper/HtmlDumper.php) is registered, so it may break the display as plain text.

zonuexe:
We can transition the default dumper from `var_dump()` to `Symfony\VarDumper`. However, it was not included in this proposal to keep compatibility.

staabm:
I guess a PR for this feature would be welcome

PR #510, opened 2017-06-24
Impl PlainTextHandler::dump()

zonuexe:
refs #509

### DIFF
--- a/src/Whoops/Handler/PlainTextHandler.php
+++ b/src/Whoops/Handler/PlainTextHandler.php
@@ -27,6 +27,11 @@ class PlainTextHandler extends Handler
     protected $logger;
 
     /**
+     * @var callable
+     */
+    protected $dumper;
+
+    /**
      * @var bool
      */
     private $addTraceToOutput = true;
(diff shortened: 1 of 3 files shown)


zonuexe:
@denis-sokolov 
Thank you for reviewing!

>  I assume you made it public for testability, but you can either use Reflection to workaround that, or at least prepend the method name with an underscore.

Yes.  I made this method public as I could not find a test helper for private methods in this project.
I applied a simple patch, but I can rewrite this as reflection.

zonuexe:
@denis-sokolov 
I pushed the method `protected` and the test using `ReflectionMethod`.
Which would you choose?

denis-sokolov:
Beautiful, I love it as is. Thank you for an exemplary PR!

src/Whoops/Handler/PlainTextHandler.php
```diff
@@ -211,6 +227,21 @@ private function getFrameArgsOutput(Frame $frame, $line)
...
 
     /**
+     * Dump variable.
+     *
+     * @param mixed $var
+     * @return void
+     */
+    protected function dump($var)
+    {
+        if ($this->dumper) {
```
prisis: is it not better to add a is_callable check her?

denis-sokolov: It can be argued both ways. I prefer not having the check, as if the dumper has been set and it is now not callable, I would rather draw attention to it immediately, and not silently fallback on the other option, making people wonder why the set dumper does not activate.

staabm: agree that a check would hide a programmer mistake

PR #512, opened 2017-06-30
Don't hide horizontal panel overflow

davispuh:
We need scrollbar because otherwise when DataTable row is too long it's cut and not visible

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -13,7 +13,6 @@ body {
   }
 
 .panel {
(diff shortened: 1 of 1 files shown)


staabm:
screenshot before/after would be great.

davispuh:
Before (no horizontal scrollbar)
attels

It doesn't happen normally, but when adding data with DataTable which uses `<pre>data</pre>` (Symphony `VarDumper`) then it's not wrapped to next line so it gets cut and there's no horizontal scrollbar.

denis-sokolov:

PR #513, opened 2017-07-03
Add open referenced files support for VSCode

tradzero:
add support for VSCode :)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -95,6 +95,7 @@ class PrettyPageHandler extends Handler
         "macvim"   => "mvim://open/?url=file://%file&line=%line",
         "phpstorm" => "phpstorm://open?file=%file&line=%line",
         "idea"     => "idea://open?file=%file&line=%line",
+        'vscode'   => "vscode://file/%file:%line",
```
denis-sokolov: Please use consistent quote style with the code around. In this case the `vscode` part should use double quotes.

tradzero:
added and fixed, thanks

denis-sokolov:

PR #516, opened 2017-08-02
safe unnecessary php_sapi_name() func call.

staabm:
use builtin constant PHP_SAPI instead.


denis-sokolov:
Indeed. We also use the constant in `Util.php`.

PR #517, opened 2017-08-02
Make PrettyPageHandler more easily extendable

juliangut:
Extending from PrettyPageHandler is a real pain: almost everything is private (nor everything should be protected neither) and most of the actual work is done in "handle" method so if I want to extend the class I've to completely copy/and/rewrite the whole method, even for the simplest of the modifications, with the problems this carries with future modifications of the original. This violates the O in SOLID

My use case: I need to filter certain frames out of the stack, and the corresponding exception code as well. To accomplish this I've just extracted "frames" and "code" collection into protected methods that can be overridden in children

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -176,27 +176,8 @@ class PrettyPageHandler extends Handler
         }
 
         $inspector = $this->getInspector();
-        $frames    = $inspector->getFrames();
-
-        $code = $inspector->getException()->getCode();
-
-        if ($inspector->getException() instanceof \ErrorException) {
-            // ErrorExceptions wrap the php-error types within the "severity" property
-            $code = Misc::translateErrorCode($inspector->getException()->getSeverity());
-        }
-
-        // Detect frames that belong to the application.
-        if ($this->applicationPaths) {
-            /* @var \Whoops\Exception\Frame $frame */
-            foreach ($frames as $frame) {
-                foreach ($this->applicationPaths as $path) {
-                    if (substr($frame->getFile(), 0, strlen($path)) === $path) {
-                        $frame->setApplication(true);
-                        break;
-                    }
-                }
-            }
-        }
+        $frames = $this->getFrames();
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -666,7 +691,7 @@ public function blacklist($superGlobalName, $key) {
      * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
      * @return array $values without sensitive data
      */
-    private function masked(array $superGlobal, $superGlobalName) {
+    private function getMasked(array $superGlobal, $superGlobalName) {
```
denis-sokolov: Why is this rename bundled with this commit? Note that the method is still private. Did you mean to make it protected as well?

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -252,9 +233,12 @@ public function handle()
 
         // Add extra entries list of data tables:
         // @todo: Consolidate addDataTable and addDataTableCallback
-        $extraTables = array_map(function ($table) use ($inspector) {
-            return $table instanceof \Closure ? $table($inspector) : $table;
-        }, $this->getDataTables());
+        $extraTables = array_map(
```
denis-sokolov: I tend to avoid reformatting in the same commit as meaningful changes, as to keep the diffs more meaningful.

juliangut: Sorry, I tend to reformat right when I see this kind of things

src/Whoops/Handler/PrettyPageHandler.php
juliangut: I toyed with the idea of making it protected but I think this is the kind of method that really is meant to be private

juliangut:
I'll update this evening for sure

juliangut:
Cleaner PR!

denis-sokolov:

juliangut:
Quick question: next release has a planed date?

denis-sokolov:
2.1.10 released.

Issue #409, opened 2016-04-17
not work in phpstorm


fotobank:
Hey. sources do not open the file in phpstorm:

use Exception;
use Illuminate\Http\Response;
use Illuminate\Validation\ValidationException;
use Illuminate\Auth\Access\AuthorizationException;
use Illuminate\Database\Eloquent\ModelNotFoundException;
use Symfony\Component\HttpKernel\Exception\HttpException;
use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
use Whoops\Handler\PrettyPageHandler;
use Whoops\Run;

at class Handler:
       protected function renderExceptionWithWhoops(Exception $e)
    {
        $whoops = new Run;
        $handler = new PrettyPageHandler();
        $handler->setEditor('phpstorm');
        $whoops->pushHandler($handler);
        return new Response(
            $whoops->handleException($e),
            $e->getStatusCode(),
            $e->getHeaders()
        );
    }

What's my mistake?

kavin-90:
Same here phpstrom does not detect whoops classes

Rarst:
What platform are you on? If I got that right `phpstorm://` protocol only works on Macs.

On Windows I have been using [RemoteCall](https://plugins.jetbrains.com/plugin/6027?pr=) plugin for this for a long time (to configure Whoops for it `handler->addEditor( 'phpstorm', 'http://localhost:8091?message=%file:%line' );`).

kavin-90:
phpstorm 2016.1

helhum:
I can confirm that it works nicely with PhpStorm 2016.1 on OSX 10.11

denis-sokolov:
I’m sorry, this does not seem to be a clearly described issue. Given your description, it is impossible to figure out what went wrong. Perhaps you should first try other avenues of action, such as investigate the problem further, ask for help on support forums. If you have new findings, by all means, feel free to add them here and the issue might be reopened.

decadence:
Same here. Looks like phpstorm:// protocol only works on Mac.
Can we change URL for supporting all platforms?
http://develar.org/idea-rest-api/#api-Platform-file
@denis-sokolov

PR #521, opened 2017-08-09
Update Open Files In An Editor.md

decadence:
Note for `phpstorm` editor because `phpstorm://` is not working on Windows and Linux.
https://youtrack.jetbrains.com/issue/IDEA-65879#comment=27-801082
https://github.com/filp/whoops/issues/409#issuecomment-217416412
At least it didn't work for me on Windows 10 and latest PhpStorm.

Not sure why you use `phpstorm://` for `phpstorm` editor because
```php
[
    "url" => "http://localhost:63342/api/file/{$file}:{$line}",
]
```
is most likely working for all platforms.


denis-sokolov:

PR #524, opened 2017-08-22
Bugfix/stacktrace is missing for haltable exceptions

Vmak11:
I noticed I was missing a stack trace for a FatalThrowableError Exception that should have had one and found it was caused by returning an empty array for the stack trace due to a commit adding ability to get stack trace from xdebug for fatal errors:
https://github.com/filp/whoops/commit/c7ab11811ab4a68d8137793e455984bfc2207073

### DIFF
--- a/src/Whoops/Exception/Inspector.php
+++ b/src/Whoops/Exception/Inspector.php
@@ -189,7 +189,7 @@ class Inspector
      *
      * If xdebug is installed
      *
-     * @param  \Throwable $exception
+     * @param  \Throwable $e
      * @return array
      */
     protected function getTrace($e)
--- a/src/Whoops/Exception/Inspector.php
+++ b/src/Whoops/Exception/Inspector.php
@@ -194,27 +194,16 @@ class Inspector
      */
     protected function getTrace($e)
     {
-        $traces = $e->getTrace();
-
-        // Get trace from xdebug if enabled, failure exceptions only trace to the shutdown handler by default
-        if (!$e instanceof \ErrorException) {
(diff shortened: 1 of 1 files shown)


src/Whoops/Exception/Inspector.php
```diff
@@ -189,32 +189,32 @@ public function getFrames()
      *
      * If xdebug is installed
      *
-     * @param  \Throwable $exception
+     * @param  \Throwable $e
      * @return array
      */
-    protected function getTrace($e)
+    protected function getTrace(\Throwable $e)
```
prisis: you cant use \Throwable her it is only available in php 7.0

staabm:
Could you add a unit test?

Vmak11:
@staabm 

Do you have any suggestions about how to unit test this as I'm not very familiar with Xdebug as well as setting up unit tests that require extensions to be installed.

staabm:
Add a unit test which reproduces the issue beeing fixed, add a "skip-condition" to not run it when xdebug is not loaded and assert via try/catch that the thrown exception will have a proper Stacktrace when provided to `getInspector()`[1]

[1] https://github.com/filp/whoops/blob/5c1d43f63f26599de3a07253a371db374752b006/docs/API%20Documentation.md#methods

denis-sokolov:
Would the following change achieve the same goal?

https://github.com/filp/whoops/compare/get-trace-fix-524

denis-sokolov:
Hopefully fixed in 51c2de8.

PR #525, opened 2017-09-06
added docs for Handler constants

staabm:
inspired by docs within `Run->handleException()`

### DIFF
--- a/src/Whoops/Handler/Handler.php
+++ b/src/Whoops/Handler/Handler.php
@@ -14,13 +14,20 @@ use Whoops\RunInterface;
  */
 abstract class Handler implements HandlerInterface
 {
-    /**
-     * Return constants that can be returned from Handler::handle
-     * to message the handler walker.
+    /*
+     Return constants that can be returned from Handler::handle
+     to message the handler walker.
      */
     const DONE         = 0x10; // returning this is optional, only exists for
                                // semantic purposes
+    /**
+     * The Handler has handled the Throwable in some way, and wishes to skip any other Handler.
+     * Execution will continue.
+     */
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/Handler.php
```diff
@@ -14,13 +14,20 @@
  */
 abstract class Handler implements HandlerInterface
 {
-    /**
-     * Return constants that can be returned from Handler::handle
-     * to message the handler walker.
+    /*
```
staabm: this comment was not related to the following `DONE` constant, therefore de-phpdoc'ed as IDE help for DONE would be confusing otherwise.

src/Whoops/Handler/Handler.php
```diff
@@ -14,13 +14,20 @@
...
+     Return constants that can be returned from Handler::handle
+     to message the handler walker.
      */
     const DONE         = 0x10; // returning this is optional, only exists for
                                // semantic purposes
+    /**
+     * The Handler has handled the Throwable in some way, and wishes to skip any other Handler.
+     * Execution will continue.
+     */
     const LAST_HANDLER = 0x20;
```
staabm: this constant somehow confused me.

I am in the process of extending the PrettyPageHandler and actually want to ignore some constants which get thrown.
"Ignore" in this context means I dont want to handle it and I dont want whoops to generate output.

I could not find a way yet how this can be achieved

staabm: ohh my problem is that Exception cannot be thrown, catched by the exception handler and control-flow goes on... this only works with errors/warnings/infos.

PR #530, opened 2017-10-01
Prevent hiding content file and/or arguments

### LINKED ISSUE #529: Css hide content
## What am I asking for

**.frame-code div hide messages** and limit content to 23 lines or 345px (for code and args/comments)

## Detailed Description

I need to display all the arguments shown in the frame-code, and the fact that i can expand the content on demand by clicking on the arrow is a good thing, but even if it is expanded, a large part of my arguments containing tables also expandable (with the arrow link) is invisible.
To see the message, I always have to manually change css value with browser inspector

## Possible Implementation

- 1 - **Deleting height limit** [line](https://github.com/filp/whoops/blob/master/src/Whoops/Resources/css/whoops.base.css#L231) in [whoops.base.css](https://github.com/filp/whoops/blob/master/src/Whoops/Resources/css/whoops.base.css) solve the problem, 

We don't risk to have big huge page cause by default 
- code view limite the number of line to [40](https://github.com/filp/whoops/blob/master/src/Whoops/Resources/views/frame_code.html.php#L24) `$range = $frame->getFileLines($line - 20, 40);`
- arguments view collapse array content

```
    .code-block {
      /* max-height: 345px; <-- remove this line */
      overflow: hidden;
      padding: 10px;
      margin: 0;
      border-radius: 6px;
      box-shadow: 0 3px 0 rgba(0, 0, 0, .05),
                  0 10px 30px rgba(0, 0, 0, .05),
                  inset 0 0 1px 0 rgba(255, 255, 255, .07);
      -moz-tab-size: 4;
      -o-tab-size: 4;
      tab-size: 4;
      }
```

- 2 - Or **change overflow value from hidden to auto**, but the risk is to display multiple scrool-bar (for each part), and i'm not sure that's a good ux solution
```
    .code-block {
      max-height: 345px; 
      overflow: auto; /* <-- change this value */
      padding: 10px;
      margin: 0;
      border-radius: 6px;
      box-shadow: 0 3px 0 rgba(0, 0, 0, .05),
                  0 10px 30px rgba(0, 0, 0, .05),
                  inset 0 0 1px 0 rgba(255, 255, 255, .07);
      -moz-tab-size: 4;
      -o-tab-size: 4;
      tab-size: 4;
      }
```

Benfarhat:
Fix #529 
limiting height to 345px for .frame-code div only display 23 line.
deleting max-height attributepermit to show all the content
Note that by default:
- code view limite the number of line to 40 $range = $frame->getFileLines($line - 20, 40);
- arguments view always collapse array content

Overview to the problem:

fix529

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -228,7 +228,6 @@ header {
     }
 
     .code-block {
-      max-height: 345px;
       overflow: hidden;
       padding: 10px;
       margin: 0;


staabm:
How does it look when your patch was applied? I dont get which problem this change fixes in the provided screenshot

Benfarhat:
Before: The **content** (code and expanded Arguments) **is hidden** caused by the max-height (345px)
Code is cutted at **line 54** and arguments at **#languages** attribute
beforepatch

After:
afterpatch

prisis:
How about a scrollbar ?

staabm:
I ran into the same issue today. tried your fix and all the arguments are rendered properly.

only problem I got: after removing this max-height whoops renders a lot more code lines (which seem to be clipped before).

I guess a scrollbar for the args instead would be good.

Benfarhat:
@staabm if you see in the second screenshot, we already have 2 scrollbars, if you add 2 others (one for the content and another one for the arguments) i'm not sure it will result in a good UX. 

> For the code: [you already have a limit of **40 lines** ](https://github.com/filp/whoops/blob/5f4bc2f496a2a63535e6bb39dcd325af557b4f29/src/Whoops/Resources/views/frame_code.html.php#L24)

> For the arguments: **arguments are visible and expanded only if you click on arrow**.

staabm:
Good points. Maybe we should reduce number of context lines in php then

Benfarhat:
@staabm if you think reducing lines is necessary, why not using a sort of "show more" button or link, and if you use a javascript/jquery script, you don't have to change TemplateHelper.php's code, but simply add a particular class (for example ".showmore") [to theses lines](https://github.com/filp/whoops/blob/master/src/Whoops/Resources/js/whoops.base.js#L37) and load the script

I make what i mean [here](https://github.com/Benfarhat/simple-show-more-less-lines)

Demo is [here ](https://benfarhat.github.io/simple-show-more-less-lines/index.html) (fully customizable)

PR #531, opened 2017-10-01
Removed type hint from masked function

zf2timo:
This type hint cause an error, when the session contains an object: 

> Fatal error: Uncaught TypeError: Argument 1 passed to Whoops\Handler\PrettyPageHandler::masked() must be of the type array, object given, called in /var/www/html/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php on line 224 and defined in /var/www/html/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php on line 691

I tried to write a unit test for this error, but this always produced an error in the `handle` function.
When a additional test is needed, guide me how to execute the `handle` function in a unit test, please .

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -688,7 +688,7 @@ class PrettyPageHandler extends Handler
      * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
      * @return array $values without sensitive data
      */
-    private function masked(array $superGlobal, $superGlobalName) {
+    private function masked($superGlobal, $superGlobalName) {
         $blacklisted = $this->blacklist[$superGlobalName];
 
         $values = $superGlobal;


staabm:
Whats your Motivation for this change?

zf2timo:
I use the `Zend\Session\Storage\SessionStrage` as my session storage. When an error occurs, I do not see an error in Whoops. Instead the an xdebug stack trace is displayed with this message

> Fatal error: Uncaught TypeError: Argument 1 passed to Whoops\Handler\PrettyPageHandler::masked() must be of the type array, object given, called in /var/www/html/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php on line 224 and defined in /var/www/html/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php on line 691

zf2timo:
I created a test script to show you want is going wrong. But I wasn't able to reproduce the error. So I think there is something wrong in my project. 
Will close this for now.

PR #536, opened 2017-10-09
Update whoops.base.js

masterwto:
for using in some off-internet networks, removing this speeds up page rendering.


staabm:
I guess https://github.com/filp/whoops/pull/399 is solving your problem.

As is this change doesnt make much sense as is.

denis-sokolov:
Thanks for your suggestion, as @staabm says, it is more beneficial for us to focus on #399 instead.

PR #540, opened 2017-10-10
Support for wildcard blacklist to blacklist a value from all superGlobals

ejunker:
I had a lot of repetition between my `_SERVER` and `_ENV` blacklist values so I figured it would be nice if there was support for the `*` wildcard. Any key blacklisted with the super global name of `*` will be blacklisted for all super globals.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -70,6 +70,7 @@ class PrettyPageHandler extends Handler
         '_SESSION' => [],
         '_SERVER' => [],
         '_ENV' => [],
+        '*' => [],
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


staabm:
Please add a test

staabm:
Default value for * missing at the class-property

ejunker:
Test added and a default value was added for `*` to the `$blacklist` property

denis-sokolov:
Thank you for your proposal. You can avoid repetition with a loop without the need to add new special syntax to Whoops. I hope this solves your use case:

```php
// Suggested
$handler->blacklist('*', 'API_KEY');

// Current
foreach(['_ENV', '_GET', '_POST', '_COOKIE', '_SERVER', '_SESSION'] as $global) {
  $handler->blacklist($global, 'API_KEY');
}
```

PR #542, opened 2017-10-13
Replace the file root path if it's different. 

mathieutu:
Hi!
Thanks for your work on this amazing package!

A lot of developers work with docker containers or virtual machines to run their web servers locally.
So when we see a Pretty Page, all the paths are wrong, and it forbids some features of Whoops.

This PR allows us to use directly the `\Whoops\Util\TemplateHelper::$applicationRootPath` to modify the display (**and editor link!**) of file's path.

So, you just have to: 
```php
$prettyHandler->setApplicationRootPath('/Users/mathieutu/Projects/MyProject');
```
to go from:
image

to:
image

In my opinion it could be really useful to allow people to customize really quickly their experience ! 

What do you think ?

### DIFF
--- a/src/Whoops/Resources/views/frame_code.html.php
+++ b/src/Whoops/Resources/views/frame_code.html.php
@@ -6,7 +6,7 @@
     <?php $line = $frame->getLine(); ?>
       <div class="frame-code <?php echo ($i == 0 ) ? 'active' : '' ?>" id="frame-code-<?php echo $i ?>">
         <div class="frame-file">
-          <?php $filePath = $frame->getFile(); ?>
+          <?php $filePath = $tpl->replaceFileRootPath($frame->getFile()); ?>
           <?php if ($filePath && $editorHref = $handler->getEditorHref($filePath, (int) $line)): ?>
             Open:
             <a href="<?php echo $editorHref ?>" class="editor-link"<?php echo ($handler->getEditorAjax($filePath, (int) $line) ? ' data-ajax' : '') ?>>
--- a/src/Whoops/Resources/views/frame_list.html.php
+++ b/src/Whoops/Resources/views/frame_list.html.php
@@ -10,7 +10,7 @@
       </div>
 
     <div class="frame-file">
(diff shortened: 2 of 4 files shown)


staabm:
Hi!
why does the existing `$prettyPageHandler->setApplicationRootPath()` not work for you?

mathieutu:
Hi @staabm!
Hum.. I'm not sure to understand your question, as this is exactly the feature I'm implementing. 😅 
Maybe I'm totally wrong but you can't do that for now. The `setApplicationRootPath` doesn't change the path of files at all.

staabm:
the PrettyPageHandler already has a method to set the ApplicationRootPath. why does this not work for your.. whats the problem with it?

I dont understand right now which problem you are trying to solve... (because it sounds like this is already solved with the existing APIs)

mathieutu:
If you read the code, you will see that for now, the `applicationRootPath` isn't used at all for this kind of thing. It's only used for the `shorten` method.

If you try to reproduce my screenshots with it, the only thing you will see is that you have broken the shorten feature ^^ .

Exact same application code, but with woops 2.1.10:
image

(See at the path at the top, and the paths in the frames.)

denis-sokolov:
I understand the value you have in the proposed change, and I value a tidy PR, thank you.

That said, I have to politely reject your proposal. The file path displayed is the one true path. That the same files are accessible outside of the container with a different path is outside the scope of Whoops’ responsibility.

As for the editor link, here you have a stronger point. But for that we already have `PrettyPageHandler::setEditor(function ($file, $line) { return string })` that allows you to customize the path in any way you need for the editor to open the files.

Please understand that the value of every feature is offset by its cost on the maintaince of the project, and in this case I think your use case can be handled well enough without adding functionality to the project.

Still, thank you for the time you took with this PR and if any part of my analysis is imprecise, please feel free to comment and continue the discussion.

mathieutu:
ok, it's too bad for the many developers who run in virtual machines or containers, but I understand your point of view.

Thank you for this complete answer.

PR #545, opened 2017-10-29
Replace print_r() with var_dump() to skip out of memory limits

jissereitsma:


src/Whoops/Util/TemplateHelper.php
```diff
@@ -183,7 +183,7 @@ public function dump($value)
             return $output;
         }
 
-        return print_r($value, true);
+        return var_dump($value, true);
     }
```
staabm: var-dump has no 2nd parameter which makes it return the value as a string.

Also I dont think that var-dump uses different memory then any other funtction

src/Whoops/Util/TemplateHelper.php
```diff
@@ -183,7 +183,7 @@ public function dump($value)
             return $output;
         }
 
-        return print_r($value, true);
+        return var_dump($value);
```
staabm: In case you want return the content from var-dump you need a output buffer to capture it

jissereitsma: I'm not sure where my mind was. `var_dump` should be `var_export`. Even though it works in Whoops, the signature of the method would not make sense.

src/Whoops/Util/TemplateHelper.php
JefferyHus: Just wondering but isn't `var_export` fails in a recursive array?

jissereitsma: Not that I know of and I didn't find anything of this liking. The official PHP documentation mentions nothing of the liking. Some user comments refer to this, but those remarks seem to relate to older PHP versions. And that's precisely the issue here: `print_r` is known to misbehave. `var_export` and `var_dump` behave slightly better ...

denis-sokolov:
I value the intent to improve this, thank you.

However, displaying data that is huge to cause memory issues is barely useful for the users, so that concern is not a priority in my opinion. And when choosing between `print_r` and `var_export`, I would rather rely on their meaning per the documentation: “human-readable information” vs “parsable string representation”, and here I choose human-readable information.

Perhaps you should disable the table output or use the Symfony dumper instead?

jissereitsma:
@denis-sokolov I did not know the flexibility of Whoops yet. I simply added `symfony/var-dumper` to my own dependencies, and this bypasses the entire issue. Thanks!

PR #547, opened 2017-11-06
Added support to Mockery 1

gabriel-caruso:
Updated `mockery/mockery` to support version [`1.0`](https://github.com/mockery/mockery/releases/tag/1.0)


denis-sokolov:

In general upgrading dependencies is valuable. But in this case there is no particular hurry right now to upgrade our developer tools, and this change has no effect on our users.

PR #548, opened 2017-11-17
BUGFIX: Make application frame selectable

DanielSiepmann:
It was not possible to activate application frame.
Now the condition is fixed and also css class is fixed.

Therefore one can select to see only application frames.


denis-sokolov:

PR #553, opened 2017-12-12
Support PHPUnit 6

carusogabriel:


denis-sokolov:
It’s good to ugprade, but we don’t need three different versions of PHPUnit. It is my understanding, we have two to support PHP 5.5 and PHP 7. If you upgrade to PHPUnit 6, perhaps we can drop PHPUnit 5?

carusogabriel:
Yes, we can have:
- `PHPUnit 4` for `PHP` versions `5.5` and `5.6`;
- `PHPUnit 6` for `PHP` versions `7.0`, `7.1` and `7.2`.

What do you think?

denis-sokolov:
I support this fully!

carusogabriel:
Something is wrong with `PHP 5.6` tests. Gonna search.

carusogabriel:
@denis-sokolov After remove PHPUnit 5 the tests fail. Should we keep it?

PR #554, opened 2017-12-12
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertStringEqualsFile` when comparing a `string` and a file;
- `assertFalse` instead of comparisons with `false` keyword;
- `assertTrue` instead of comparisons with `true` keyword;
- `assertArrayHasKey` instead of `isset` function;
- `assertNotSame` instead of strict comparisons `===`.

### DIFF
--- a/tests/Whoops/Exception/FrameCollectionTest.php
+++ b/tests/Whoops/Exception/FrameCollectionTest.php
@@ -66,7 +66,7 @@ class FrameCollectionTest extends TestCase
     public function testArrayAccessExists()
     {
         $collection = $this->getFrameCollectionInstance();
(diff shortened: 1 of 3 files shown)


denis-sokolov:
Yeah, these are good. Could you please confirm all of the methods work in PHPUnit 4.8 we have in our composer.json?

carusogabriel:
Yes, they all work in `PHPUnit 4.8` according to [its manual](https://phpunit.de/manual/4.8/pt_br/appendixes.assertions.html#appendixes.assertions).

denis-sokolov:
Thank you!

PR #558, opened 2018-01-08
Changed key masking to display 'BLACKLISTED'

brysonreece:
The reason for this is simply in the interest of security; by repeating '*' for each key's characters it makes it the slightest bit easier to tailor an attack by knowing each key's length. By displaying 'BLACKLISTED', you're both continuing to hide the specified superglobal key and mitigating a possible attack exploitation.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -698,7 +698,7 @@ class PrettyPageHandler extends Handler
         $values = $superGlobal;
         foreach ($blacklisted as $key) {
             if (isset($superGlobal[$key])) {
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thanks for the contribution! I agree with the security value of your proposal. However, I disagree on a subjective judgement call on how much security is beneficial to us here at cost of how much of user experience.

The entire blacklisting feature is questionable from the beginning, since this is a development tool that shall never see the production front-end. I agreed to sacrifice a tiny bit of extra code compleity for a considerable security benefit. This proposal, however, sacrififices a bunch of aesthetics and UX for a very small security benefit.

Even though we did not merge this in, I thank you for creating the suggestion. Integrating software with other people is always work, it’s debate and discussions, so please don’t be discouraged to try other ideas later on!

PR #560, opened 2018-01-10
Fixes from PHPStan

carusogabriel:
Some fixes founded by [PHPStan](https://github.com/phpstan/phpstan) :man_technologist:

### DIFF
--- a/src/Whoops/Exception/FrameCollection.php
+++ b/src/Whoops/Exception/FrameCollection.php
@@ -20,7 +20,7 @@ use UnexpectedValueException;
 class FrameCollection implements ArrayAccess, IteratorAggregate, Serializable, Countable
 {
     /**
-     * @var array[]
+     * @var Frame[]
      */
     private $frames;
 
@@ -181,7 +181,7 @@ class FrameCollection implements ArrayAccess, IteratorAggregate, Serializable, C
      * Gets the innermost part of stack trace that is not the same as that of outer exception
      *
      * @param  FrameCollection $parentFrames Outer exception frames to compare tail against
-     * @return array
+     * @return Frame[]
      */
     public function topDiff(FrameCollection $parentFrames)
     {
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -388,7 +388,7 @@ class PrettyPageHandler extends Handler
      *       return "http://stackoverflow.com";
      *   });
      * @param string $identifier
-     * @param string|\Closure $resolver
+     * @param string|callable $resolver
      */
     public function addEditor($identifier, $resolver)
     {
--- a/src/Whoops/Handler/XmlResponseHandler.php
+++ b/src/Whoops/Handler/XmlResponseHandler.php
@@ -7,6 +7,7 @@
 namespace Whoops\Handler;
 
 use SimpleXMLElement;
+use Traversable;
 use Whoops\Exception\Formatter;
 
 /**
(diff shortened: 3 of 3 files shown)


src/Whoops/Exception/FrameCollection.php
```diff
@@ -181,7 +181,7 @@ public function prependFrames(array $frames)
      * Gets the innermost part of stack trace that is not the same as that of outer exception
      *
      * @param  FrameCollection $parentFrames Outer exception frames to compare tail against
-     * @return Frame[]
+     * @return array
```
staabm: should not be changed, the more specific one is better

src/Whoops/Exception/Frame.php
```diff
@@ -206,8 +206,8 @@ public function getFileLines($start = 0, $length = null)
 
             // Get a subset of lines from $start to $end
             if ($length !== null) {
-                $start  = (int) $start;
-                $length = (int) $length;
+                $start  = $start;
```
staabm: I guess dropping all the casts could be considered a BC break. only because of a phpdoc comment saying it should be a int does not necessarily mean everyone out their used it only with ints.

src/Whoops/Exception/FrameCollection.php
carusogabriel: What about changing the `$frame` property to `Frame[]` instead of `array`?
https://github.com/filp/whoops/blob/4d363e3711f0dcc6c770b94e24fce80777bc0308/src/Whoops/Exception/FrameCollection.php#L22-L25

src/Whoops/Exception/Frame.php
carusogabriel: Maybe we should add tests instead? Or extend the `phpDocs`?

src/Whoops/Exception/FrameCollection.php
staabm: not sure how this is releated to the change mentioned in this diff, but makes sense.

src/Whoops/Handler/XmlResponseHandler.php
```diff
@@ -67,7 +67,7 @@ public function contentType()
      */
     private static function addDataToNode(\SimpleXMLElement $node, $data)
     {
-        assert(is_array($data) || $data instanceof Traversable);
+        assert(is_array($data) || $data instanceof \Traversable);
```
GrahamCampbell: Stylistically preferable to import this class instead.

tests/Whoops/RunTest.php
```diff
@@ -40,7 +40,7 @@ protected function getException($message = null)
     }
 
     /**
-     * @return Handler
+     * @return \Mockery\MockInterface
```
GrahamCampbell: Hmm, the handler type was used on purpose to indicate the interface.

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -388,7 +388,7 @@ public function handleUnconditionally($value = null)
      *       return "http://stackoverflow.com";
      *   });
      * @param string $identifier
-     * @param string $resolver
+     * @param string|\Closure $resolver
```
GrahamCampbell: `callable`?

carusogabriel: Yes, I just read the method description now, sorry :sweat_smile:

src/Whoops/Handler/XmlResponseHandler.php
carusogabriel: Importing...

GrahamCampbell: Yeh, I'd say so.

PR #565, opened 2018-02-21
Add inlcude RunInterface.php

junaexp:
work for codeignither


staabm:
autoloading should take care of this include. when you need this line at this spot your autoload config seems to be borked.

denis-sokolov:
As @staabm said, autoloading should be handling this already. Try to figure out what’s wrong with your setup and thanks for trying to help out!

PR #570, opened 2018-03-28
Fix security

Exitialis:
For security reasons. If someone forget disable debug mode, anyone can see length of passwords and the brute force it

Before: 
image

After:
image

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -692,7 +692,7 @@ class PrettyPageHandler extends Handler
      * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
      * @return array $values without sensitive data
      */
-    private function masked(array $superGlobal, $superGlobalName, int $length = 20)
+    private function masked(array $superGlobal, $superGlobalName, $length = 20)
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -692,14 +692,14 @@ public function blacklist($superGlobalName, $key)
      * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
      * @return array $values without sensitive data
      */
-    private function masked(array $superGlobal, $superGlobalName)
+    private function masked(array $superGlobal, $superGlobalName, $length = 20)
```
prisis: No need for this custom variable, nobody gone change this

BenjaminHoegh:
in what way is this a security fix? You turn off error messages when going live. Whoops is only for the developer to get a better debugger

denis-sokolov:
This is a good suggestion, but the minor security benefit it brings needs to be weight against the minor developer inconvenience when reviewing their values. It’s a judgement call, and because Whoops is not intended for production, I will prioritize the developer’s experience and choose not to include this feature. But thanks for suggesting and raising this question!

PR #572, opened 2018-04-19
Updated default header size to big debug strings

uonick:
Hi! 

**Many** errors in `Laravel` have a **huge** number of rows and resize action (on hover) in div is **very** slow.

To speed up the development and ease the work of programmers, I suggest immediately making a window above!

### before

screen shot 2018-04-19 at 17 59 18

### after

screen shot 2018-04-19 at 17 59 32

### DIFF
--- a/src/Whoops/Resources/css/whoops.base.css
+++ b/src/Whoops/Resources/css/whoops.base.css
@@ -38,7 +38,7 @@ header {
   box-sizing: border-box;
   background-color: #2a2a2a;
   padding: 35px 40px;
-  max-height: 180px;
+  max-height: 500px;
   overflow: hidden;
   transition: 0.5s;
 }


src/Whoops/Resources/css/whoops.base.css
```diff
@@ -38,7 +38,7 @@ header {
   box-sizing: border-box;
   background-color: #2a2a2a;
   padding: 35px 40px;
-  max-height: 180px;
+  max-height: 500px;
```
staabm: Maybe this should be relative to the viewportheight with `vh` units?

uonick: Yes, sure

uonick: `vh` - does not give the necessary result...

denis-sokolov:
This is a difficult one. Sometimes longer exception messages are more useful and sometimes they’re not useful, while a stacktrace is. For now I will not include this, but I’ll be on the lookout for how to solve this better. Thanks!

PR #577, opened 2018-06-01
Universal blacklist.

finagin:
Universal blacklist a sensitive value within one of the superglobal arrays.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -63,6 +63,7 @@ class PrettyPageHandler extends Handler
      * @var array[]
      */
     private $blacklist = [
+        '*' => [],
         '_GET' => [],
         '_POST' => [],
         '_FILES' => [],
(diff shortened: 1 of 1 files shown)


barryvdh:
Yeah was looking in how it was currently tested, but couldn't find it. Not really sure how to test it standalone, as they are private.

staabm:
Ah correct. We dont bave tests for this method

SvenRtbg:
It should still be documented how to use this feature. I.e. at least change the comment for the `blacklist` method to mention that `*` is a wildcard for `_GET, _POST, _FILES, _COOKIES, _SESSION, _SERVER, _ENV`.

One could argue that adding wildcard entries to all these existing arrays instead of adding a new category `*` array (and merging later) is a more straightforward approach and may lead to less trouble whenever changing the `masked(array $superGlobal, $superGlobalName)` method. The current implementation does rely on the blacklist array to be _numerically_ indexed arrays of strings, so `array_merge()` will append everything in the wildcard list, and then reduce double work by applying `array_unique()`.

denis-sokolov:
I’m with @SvenRtbg, this is a good feature, but the number of questions it raises is not worth the minor benefit it brings. If you want to redesign, please don’t be dissuaded by the closing, we can discuss it further.

PR #579, opened 2018-06-13
added hide-error button

XzenTorXz:


axcsds:
cool.

staabm:
please elaborate a bit why and when this change is usefull. as is, the chance this gets merged is really low.

XzenTorXz:
its just a little helper, with the same design as the "copy" button. It will hide the fullscreen error message.

This is useful in situation when you allready have an output send to your browser, that might bring some more context to your error message. For example the good old var dump debugging f.e.

```
<?php
...
var_dump($var); //this output will not be visible after the exeption is thrown
flush();
throw new Exception("whoops");
```

PR #582, opened 2018-06-30
editor callback: allow to return empty value

gharlan:
In our case we have some virtual paths (stream wrappers), and some of them can be converted into an editor url, others can not.

So I suggest to allow returning an empty value from editor callback.


staabm:
can we get another release tagged. I dont have permission to do so on github.

denis-sokolov:
@staabm 2.2.1

Issue #586, opened 2018-07-07
Support source code preview when php codes is not executed from a file


staabm:
There are different use cases where the actual php code beeing run is not contained within a file.

In some applications some php source might be stored in e.g. database or in our concrere use-case the source is included via a php stream wrapper.

The current mechanics dont allow to render the code preview for those code frames.

I would like to request some kind of api which allows to find/fetch the actual source per frame for non-file-based source locations.

gharlan:
Maybe a possibility to set a "frame handler" callback, which will be called for each frame.
This could be used [to add comments](https://github.com/filp/whoops/blob/e79cd403fb77fc8963a99ecc30e80ddd885b3311/src/Whoops/Exception/Frame.php#L133-L144) or to set the "file" content to bypass [this](https://github.com/filp/whoops/blob/e79cd403fb77fc8963a99ecc30e80ddd885b3311/src/Whoops/Exception/Frame.php#L133-L144).

denis-sokolov:
That’s a fancy advanced feature that will see very little usage. Still, I would be open for a PR if it is carefully designed.

PR #594, opened 2018-10-18
Remove superfluous is_file() check

staabm:
this lines were added in the past to handle the case where $filePath is „Unknown“, see 
https://github.com/filp/whoops/commit/f83625d66aa85dd4a49260696930a3c3ea2c1af3

Later on a additional check for „Unknown“ was added for open_basedir related errors
https://github.com/filp/whoops/commit/1c0ffa00594622d7d1f4368de5ae2249617e5167

Therefore the lines added in the first place are no longer usefull.

Additionally there is a feature request in https://github.com/filp/whoops/issues/586 for which the mentioned lines are getting in our way.

### DIFF
--- a/src/Whoops/Exception/Frame.php
+++ b/src/Whoops/Exception/Frame.php
@@ -119,11 +119,6 @@ class Frame implements Serializable
                 return null;
             }
 
-            // Return null if the file doesn't actually exist.
-            if (!is_file($filePath)) {
-                return null;
-            }
-
             $this->fileContentsCache = file_get_contents($filePath);
         }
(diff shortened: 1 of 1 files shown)


staabm:
@gharlan could you confirm this fix?

staabm:
I just verified that this works as expected.

therefore this is ready to land.

staabm:
> Maybe we should also set the file contents cache to some special value, so we know to not keep going to the disk over and over if the file doesn't exist?

Did you ever run into a situation in which it mattered?

Issue #593, opened 2018-10-04
message from previous exceptions missing


geromueller:
Please concat the message of the previous exceptions. currently only the stack frames are collected. The messages are even more important!

denis-sokolov:
This is a good suggestion. If anyone is up to implement it, I would be open to accept a PR.

For future reference, added this to [the wiki page of possible features](https://github.com/filp/whoops/wiki/Possible-features-to-add).

denis-sokolov:
Done in #596!

PR #596, opened 2018-10-20
Display messages from previous exceptions

florianorben:
Hey!

This will show a list of all the previous exceptions that have been thrown before the current exception as described in #593.

It'll look like this
preview

References #593 .


src/Whoops/Resources/views/header.html.php
```diff
@@ -15,6 +15,25 @@
...
+
+      <?php if (count($previousMessages)): ?>
+        <div class="exc-title prev-exc-title">
+          <span class="exc-title-secondary">Previous Exceptions</span>
+        </div>
+
+        <ul>
+          <?php foreach ($previousMessages as $i => $previousMessage): ?>
+            <li>
+              <span class="prev-exc-code">(<?php echo $previousCodes[$i] ?>)</span>
```
staabm: In the top-level exception we used to have the exc.-behind the message

florianorben: Oops, your're right. :)
image
Changed now

staabm:
Really like this. Didnt have time to test it

florianorben:
Also, changed "Previous Exceptions" to "Previous exceptions" (lowercase "e").
Didn't update the screenshot though :)

florianorben:
@denis-sokolov Done! Was used to using typehints, but PHP 5.6 bugged with using  `\Throwable` :)
The latest HHVM build seems to have bugged because of some travis issue
```
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown
```
is there any way to trigger a new build without a push?

staabm:
I guess open+close PR will work

florianorben:
Didn't work sadly :( Changing a random comment and a new push did it.
All builds green now :)

denis-sokolov:
CIs usually have a button to restart the build. In this case only I have access to it if I am logged in.

Re-pushing works in all cases, though. I usually do a `git commit --amend`, don’t change the commit message, and `git push -f`. An amend updates the timestamp which modifies the hash which makes it a new commit.

Thank you for a great contribution!

Issue #598, opened 2018-10-21
file_get_contents([internal]): failed to open stream: No such file or directory


lcharette:
We're encountering a `file_get_contents([internal]): failed to open stream: No such file or directory` error in our implementation of whoops. After some digging, it goes back to the changes made in https://github.com/filp/whoops/commit/4798f07fc8987457073a2184cbef0afdc4d54fdb.

The `[internal]` part is returned by the [`Inspector`](https://github.com/filp/whoops/blob/master/src/Whoops/Exception/Inspector.php#L181). Removing the superfluous `is_file()` check doesn't really make sense if a file can be set as `[internal]` and won't be catched by the other checks. 

A proposed fix would be to revert changes made in https://github.com/filp/whoops/commit/4798f07fc8987457073a2184cbef0afdc4d54fdb.

staabm:
Can you describehow to reproduce this problem?

lcharette:
I managed to replicate in a fresh clone of this repo by changing this line : https://github.com/filp/whoops/blob/master/tests/Whoops/Handler/PrettyPageHandlerTest.php#L21

To : 
```
$handler->handleUnconditionally(true);
```

Same happens if you change the default property to true ( `private $handleUnconditionally = true;` )

denis-sokolov:
Fixed in #599!

PR #599, opened 2018-10-22
Prevent exception from Whoops when caught exception frame is not related to real file

SilverFire:


staabm:
I guess this is a fix for https://github.com/filp/whoops/issues/598

would be great to have a unit-test covering the change.

@lcharette can you verify the fix?

SilverFire:
Yes, seems to be a fix for #598 :)

> would be great to have a unit-test covering the change.

Done

tests/Whoops/Exception/FrameTest.php
```diff
@@ -103,6 +103,20 @@ public function testGetFileContents()
...
 
+    /**
+     * @covers Whoops\Exception\Frame::getFileContents
+     * @testWith ["[internal]"]
+     *           ["Unknown"]
+     * @see https://github.com/filp/whoops/pull/599
+     */
+    public function testGetFileContentsWhenFrameIsNotRelatedToSpecificFile($fakeFilename)
+    {
+        $data  = array_merge($this->getFrameData(), ['file' => $fakeFilename]);
```
staabm: I guess we need php 5.x compat here

```suggestion
        $data  = array_merge($this->getFrameData(), array('file' => $fakeFilename));
```

lcharette: FYI, since you require php 5.5.9  or above, you should be fine.

> Following [] is supported in PHP >= 5.4:

lcharette:
Looks like it would indeed fix #598. I'll be able to test later today.

denis-sokolov:

PR #600, opened 2018-10-28
Support to prepend handler

### LINKED ISSUE #587: Add support to prepend handler
Add support to prepend handler. When using this, handler should be added before other handlers. Also, don't forget to update `RunInterface` with new function.

I found simple solution:
```php
public function prependHandler($handler)
{
    $existingHandlers = array_merge([$handler], $this->getHandlers());
    $this->clearHandlers();

    foreach ($existingHandlers as $existingHandler) {
        $this->pushHandler($existingHandler);
    }
}
```

filips123:
Add support to prepend handler. Fix #587.

### DIFF
--- a/src/Whoops/RunInterface.php
+++ b/src/Whoops/RunInterface.php
@@ -26,6 +26,15 @@ interface RunInterface
     public function pushHandler($handler);
 
     /**
+     * Prepends a handler to the begining of the stack
+     *
+     * @throws InvalidArgumentException  If argument is not callable or instance of HandlerInterface
+     * @param  Callable|HandlerInterface $handler
+     * @return Run
+     */
+    public function prependHandler($handler);
+
+    /**
      * Removes the last handler in the stack and returns it.
      * Returns null if there"s nothing else to pop.
      *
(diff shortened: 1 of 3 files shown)


src/Whoops/RunInterface.php
```diff
@@ -25,6 +25,15 @@
...
     public function pushHandler($handler);
 
+    /**
+     * Prepends a handler to the begining of the stack
+     *
+     * @throws InvalidArgumentException  If argument is not callable or instance of HandlerInterface
+     * @param  Callable|HandlerInterface $handler
+     * @return Run
+     */
+    public function prependHandler($handler);
```
staabm: Changing the interface is a BC break. We need to check whether this is a problem..

filips123: Is it OK to add new function but not update the interface?

filips123: @denis-sokolov @staabm Is it OK?

denis-sokolov:
For now #587 is closed, I’m closing this, too. If we reconsider in #587, we can revisit this.

Issue #604, opened 2018-11-11
Prevent Whoops from being exposed on non-development environment


kktsvetkov:
**TL;TR** 
Don't do anything with PrettyPageHandler if you are on a real public IP address or if the user-agent is a "bot".

Longer Version:

You can't fix _stupid_, but you can make it harder for careless developers to shoot themselves in the foot.

Earlier today I was commenting on the [dumping arguments change](https://github.com/filp/whoops/pull/387#issuecomment-437657944), and the example I gave I found  googling for ["Whoops! There was an error"](https://www.google.com/search?q=%22Whoops!%20There%20was%20an%20error%22). You've probably seen all of that already. 

Exposing your project's exceptions is bad, and it is even worse if the Whoops page shares some sensitive data such as credentials and keys/tokens. I ask was asking if the blacklist feature must cover the arguments dump as well. @jonasdt commented that Whoops is meant to be used for development environments only anyway, and I agree -- it's the developers' responsibility to set it up correctly. Nevertheless, Google is full of crawled Whoops page.

My suggestion is that add two extra checks that will block the exception rendering:

1. allow to proceed if SERVER_ADDR is a private IP (10.*, 172.16.*, 192.168.*), deny if SERVER_ADDR is public IP address. My assumption is that most development environments are done on private networks, so looking at the IP is a somewhat good way to determine what environment Whoops is running in.

1. deny rendering the exception if the user-agent is a "bot". My assumption is that even if the Whoops rendering of the exception is publicly available, at least deny the crawling bots to index it. If you look at cached Google Whoops pages, you are going to see that they have:
 - `HTTP_FROM googlebot(at)googlebot.com`
 - `HTTP_USER_AGENT Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.96 Mobile Safari/537.36 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)`
 - `HTTP_USER_AGENT Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)`

So, It's a long shot, but anyway - this is a good precaution against careless developers.

shadowhand:
> My assumption is that most development environments are done on private networks, so looking at the IP is a somewhat good way to determine what environment Whoops is running in.

[...]

PR #605, opened 2018-11-11
Added noindex http headers

staabm:
Its redundant a bit, but better be safe then sorry.

Proxies/webserver might override the http status code and therefore error pages could show up in e.g. google search index

https://www.google.de/search?q=Whoops%21+There+was+an+error

Refs https://github.com/filp/whoops/issues/604

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -300,6 +300,7 @@ final class Run implements RunInterface
 
                 // Send any headers if needed:
                 if (Misc::canSendHeaders() && $handlerContentType) {
+                    header("X-Robots-Tag: noindex, nofollow");
                     header("Content-Type: {$handlerContentType}");
                 }
             }


staabm:
Not 100% sure this header should better be sent in writeToOutputNow()

https://github.com/filp/whoops/blob/45dad4e41eff66217626029f4172b04dc4025442/src/Whoops/Run.php#L400

kktsvetkov:
Why not add `<meta name="robots" content="noindex" />` in HTML output as well ? Meta tags will be valid even after headers_sent()

staabm:
The pretty-page-handler already contains this markup:
https://github.com/filp/whoops/blob/45dad4e41eff66217626029f4172b04dc4025442/src/Whoops/Resources/views/layout.html.php

This http-header has the benefit of also working for other handlers

PR #609, opened 2018-12-03
add previous exception messages to plain text output

### LINKED ISSUE #608: PlainTextHandler does not show previous exceptions
I'm not sure if I should open two issues and before I spam this really nice project I wanted to ask the maintainers what they think about it.

The initial problem I had is that somewhere in the code an exception can be catch and replaced with another exception. Ideally the exception contains the previous exception that caused this exception.

For example an ORM catches the PDO exception and throws an ORM specific exception. I'm ok with this approach and I don't think the ORM specific exception should include the message from the PDO exception but it should have the previous exception. But with PlainTextHandler I only get the uncaught exception.

Now just because I don't wanted to bother you I wanted to extend the PlainTextHandler to lets say a DetailedPlainTextHandler. But what I really don't like is to copy paste  the exact same method just because it is private and so I'm writing here because `PlainTextHandler::getTraceOutput()` is private so I can't just extend the PlainTextHandler.

I would love to add the option `PlainTextHandler::addPreviousToOutput()` that is just adding all messages followed by the stack trace. Example output:

```
ORM\Exception: Error preparing the fetch statement for App\Model\User in file /path/to/orm.php on line 23
Before: PDOException: undefined table users in /path/to/orm.php on line 42
#1 ORM\Exception->() /path/to/orm.php:23
#2 PDO->prepare('...') /path/to/orm.php:42
#3 ORM\EntityManager->fetch('App\Model\...') /path/to/script.php:6
```

Would such a PR be accepted or is there already a similar PR or feature?

<small>p. s.: can someone please explain why `PlainTextHandler::getTraceOutput()` is private? It is nearly impossible to extend the class without having to write everything your own.</small>

tflori:
fixes #608 

Can be disabled via `PlainTextHandler::addPreviousToOutput(false)` and
adds lines between the exception and the trace containing the exception
and line of creator.

The method `Inspector::getFrames()` already appends all frames from the
previous exceptions. So there is no need in adding the trace for each
previous exception.

IMO the test tests something that it should not test: the output of the
line number. But that should be in a separate PR. I've at least added
constants for the line numbers so that you only have to change them once
for all tests.

### DIFF
--- a/src/Whoops/Handler/PlainTextHandler.php
+++ b/src/Whoops/Handler/PlainTextHandler.php
@@ -49,6 +49,11 @@ class PlainTextHandler extends Handler
     /**
      * @var bool
      */
+    private $addPreviousToOutput = true;
+
+    /**
+     * @var bool
+     */
     private $loggerOnly = false;
 
     /**
@@ -115,6 +120,21 @@ class PlainTextHandler extends Handler
     }
 
     /**
+     * Add previous exceptions to output.
+     * @param  bool|null $addPreviousToOutput
+     * @return bool|$this
+     */
+    public function addPreviousToOutput($addPreviousToOutput = null)
+    {
+        if (func_num_args() == 0) {
+            return $this->addPreviousToOutput;
+        }
+
+        $this->addPreviousToOutput = (bool) $addPreviousToOutput;
+        return $this;
+    }
+
+    /**
      * Add error trace function arguments to output.
      * Set to True for all frame args, or integer for the n first frame args.
      * @param  bool|integer|null $addTraceFunctionArgsToOutput
@@ -151,14 +171,18 @@ class PlainTextHandler extends Handler
     public function generateResponse()
     {
         $exception = $this->getException();
-        return sprintf(
-            "%s: %s in file %s on line %d%s\n",
-            get_class($exception),
-            $exception->getMessage(),
-            $exception->getFile(),
-            $exception->getLine(),
-            $this->getTraceOutput()
-        );
+        $message = $this->getExceptionOutput($exception);
+
+        if ($this->addPreviousToOutput) {
+            $previous = $exception->getPrevious();
+            while ($previous) {
+                $message .= "\nBefore: " . $this->getExceptionOutput($previous);
+                $previous = $previous->getPrevious();
+            }
+        }
+
+
+        return $message . $this->getTraceOutput() . "\n";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Whoops/Handler/PlainTextHandler.php
```diff
@@ -151,14 +171,18 @@ public function setTraceFunctionArgsOutputLimit($traceFunctionArgsOutputLimit)
...
-            $exception->getFile(),
-            $exception->getLine(),
-            $this->getTraceOutput()
-        );
+        $message = $this->getExceptionOutput($exception);
+
+        if ($this->addPreviousToOutput) {
+            $previous = $exception->getPrevious();
+            while ($previous) {
+                $message .= "\nBefore: " . $this->getExceptionOutput($previous);
```
staabm:
```suggestion
                $message .= "\nPrevious: " . $this->getExceptionOutput($previous);
```

src/Whoops/Handler/PlainTextHandler.php
```diff
@@ -284,6 +308,22 @@ private function getTraceOutput()
         return $response;
     }
 
+    /**
+     * Get the exception as plain text.
+     * @param \Throwable $exception
+     * @return string
+     */
+    protected function getExceptionOutput($exception)
```
staabm:
```suggestion
    private function getExceptionOutput($exception)
```

tflori:
I just realized how phpunit outputs the trace - they have some more line breaks:

```
Exception: Outer exception message
<trace>

Caused by
Exception: Inner exception

<trace>
```

The trace for later exceptions is just a subset of the last but dividing the output with a line break and the Caused by keyword makes it easy to find specific exceptions with `egrep '^MyException:' app.log`.

What do you think?

staabm:
Go for it.

src/Whoops/Handler/PlainTextHandler.php
```diff
@@ -46,6 +46,11 @@ class PlainTextHandler extends Handler
      */
     private $traceFunctionArgsOutputLimit = 1024;
 
+    /**
+     * @var bool
+     */
+    private $addPreviousToOutput = true;
```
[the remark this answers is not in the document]
tflori: maybe someone don't want it in the logs. or they show it to the client and want to hide it there (for example the message could contain internal information that should only be visible to developers).

src/Whoops/Handler/PlainTextHandler.php
tflori: updated with a line break and caused by information...

src/Whoops/Handler/PlainTextHandler.php
tflori: if you really insist in private methods.. done

denis-sokolov:

PR #611, opened 2018-12-08
Replace "E_ALL | E_STRICT"  with "E_ALL"

michaelKaefer:
See my more detailed explanation here: https://stackoverflow.com/questions/1638238/whats-the-point-of-e-all-e-strict-if-its-the-same-value-as-e-all/53685722#53685722


denis-sokolov:

PR #612, opened 2018-12-10
Add wildcard to masked blacklist

barryvdh:
This will make it easier to mask variables. For example: `*_KEY`, `*_PASSWORD` or just `*` to just mask every input for that superglobal.


denis-sokolov:
Thanks for a good suggestion!

Unfortunately, the masking is a very low-priority feature and this grows its complexity further to uncomfortable levels.

Instead of custom pattern language, I would accept a PR with a lambda filter function. This would also give the user more power to dynamically decide what to do.

I would also kindly ask for a pair of tests for a feature of this complexity, otherwise there is no doubt we will break it in the future. 🤷‍♂

PR #620, opened 2019-02-28
Added support for opening files in Espresso editor

olach:
This is a simple PR that enables support for opening files in the Espresso text editor. The docs are also updated.

For reference: https://help.espressoapp.com/extensions/x-espresso.html


denis-sokolov:

PR #621, opened 2019-02-28
use fqcn in example

staabm:
use full qualified class-names in example codes to make them more self-contained and also work in namespaced code


denis-sokolov:
👍

PR #624, opened 2019-04-29
File contents fix

zsilbi:
Hey!

Errors in internal source files (for example PHP extensions like PhalconPHP) cause error loops in some rare cases.

In this example, Phalcon v4 alpha (https://github.com/phalcon/cphalcon) is used and, an error is reported in an internal Zephir source file that cannot be opened via file_get_contents().

Result:
```php
Uncaught Whoops\Exception\ErrorException: file_get_contents(phalcon/Flash/Session.zep): failed to open stream: No such file or directory in...
Stack trace:
#0 [internal function]: Whoops\Run->handleError(2, 'file_get_conten...', '/site/vendor...', 122, Array)
#1 /site/vendor/filp/whoops/src/Whoops/Exception/Frame.php(122): file_get_contents('phalcon/Flash/S...')
```

Thanks,
zsilbi


zsilbi:
@denis-sokolov, I think the CI for hhvm is might be broken.
Any ideas? Shall I try again?

Thanks

src/Whoops/Exception/Frame.php
```diff
@@ -119,7 +119,11 @@ public function getFileContents()
                 return null;
             }
 
-            $this->fileContentsCache = file_get_contents($filePath);
+            try {
+                $this->fileContentsCache = file_get_contents($filePath);
```
SvenRtbg: Why not returning the result here? Otherwise `return null` in the catch block makes no sense because it'll look weird as a replacement.

zsilbi: You are right.
It also could be just an empty catch block, because `fileContentsCache` is `null` anyways.

denis-sokolov:

PR #625, opened 2019-04-30
Make call static

### LINKED ISSUE #602: ::toXML() method in Whoops/Handler/XmlResponseHandler.php
Hi, just saw something that seems odd. At whoops/src/Whoops/Handler/XmlResponseHandler.php, line 50 there is

`echo $this->toXml($response);`

but the ::toXML() method is declared as `static` at line 99:

`private static function toXml($data)`

Perhaps at line 50 it should be 

`echo self::toXml($response);`

I saw that originally ::toXml() was public and non-static, and that in commit [720648a](https://github.com/filp/whoops/commit/720648a) it was changed to `private static` and probably it was missed to convert the line 50 as well.

SvenRtbg:
Fixes #602

### DIFF
--- a/src/Whoops/Handler/XmlResponseHandler.php
+++ b/src/Whoops/Handler/XmlResponseHandler.php
@@ -47,7 +47,7 @@ class XmlResponseHandler extends Handler
             ),
         ];
 
-        echo $this->toXml($response);
+        echo static::toXml($response);
 
         return Handler::QUIT;
     }
--- a/src/Whoops/Handler/XmlResponseHandler.php
+++ b/src/Whoops/Handler/XmlResponseHandler.php
@@ -47,7 +47,7 @@ class XmlResponseHandler extends Handler
             ),
         ];
 
-        echo static::toXml($response);
+        echo self::toXml($response);
 
         return Handler::QUIT;
     }


src/Whoops/Handler/XmlResponseHandler.php
```diff
@@ -47,7 +47,7 @@ public function handle()
             ),
         ];
 
-        echo $this->toXml($response);
+        echo static::toXml($response);
```
staabm:
```suggestion
        echo self::toXml($response);
```

?

prisis: If you use self nobody can extend this function

staabm: The  function is private

SvenRtbg: I've no preference here, just wanted to close a simple issue that I came across.

SvenRtbg:
Fatal error: Uncaught Error: Keyword __HALT_COMPILER must be written in lowercase in /home/travis/.phpenv/versions/hhvm-stable/bin/composer:27

I think hhvm testing just went down the drain with Composer being incompatible in TravisCI.

denis-sokolov:

Issue #627, opened 2019-05-16
strlen() expects parameter 1 to be string, array given


andrewwindfall:
Hello,
I get the following error on any laravel error. I have recently upgraded from 5.6 to 5.8, but I cannot say for certain if the problem started after the upgrade. My composer.json looks like:

(1/1) ErrorException
strlen() expects parameter 1 to be string, array given
--
in PrettyPageHandler.php line 708

```

"require": {
        "php": ">=7.1.3",
        "fideloper/proxy": "~4.0",
        "jgrossi/corcel": "^2.7",
        "laravel/framework": "5.8.*",
        "laravel/tinker": "~1.0"
    },
    "require-dev": {
        "filp/whoops": "~2.0",
        "fzaninotto/faker": "~1.4",
        "mockery/mockery": "~1.0",
        "phpunit/phpunit": "^7.5",
        "symfony/thanks": "^1.0"
    },

```

Php info:
`php 7.2.14`

staabm:
Do you have a stacktrace and/or file/lineno of the issue/warning in question?

andrewwindfall:
Here is the full stack
```

(1/1) ErrorExceptionstrlen() expects parameter 1 to be string, array given
--
in PrettyPageHandler.php line 708
[...]

PR #628, opened 2019-05-16
Prevent „strlen() expects parameter 1 to be string, array given“ Warning

staabm:
Should fix https://github.com/filp/whoops/issues/627

The build error is not related


staabm:
Fix confirmed by the issue reporter

williamdes:
cc @denis-sokolov ?

williamdes:
Thanks @denis-sokolov 
Waiting for a new version now :)

Issue #580, opened 2018-06-23
Missing an unshift handler method


tflori:
Because handlers may stop to execute other handlers I need to set the order correctly but most important handlers are added first in my application.

I'm initializing my application and whoops should handle all errors. And yet I don't know which handlers I want to register (is it a web page and the user is allowed to see error messages - then I want to register a PrettyPage handler; or it is a json request; or it is a cli interface) but all errors should everytime appear in a log. So at this stage I want to register a plain text handler with `loggerOnly(true)`.

But when I then register a pretty page handler it quits the execution of other handlers and so the error will not be logged.

My current workaround is to register the handlers with a helper function that has a confusing naming (of course `Run` is final so not in a subclass of `Run` where it should be):

```php
/**
 * Prepends $handler to the list of handlers from whoops.
 *
 * Because whoops executes the handlers in reverse order they will run as last handler - so they are appended.
 *
 * @param callable|HandlerInterface $handler
 */
public function appendWhoopsHandler($handler)
{
    $handlers = $this->whoops->getHandlers();
    array_unshift($handlers, $handler);
    $this->whoops->clearHandlers();
    foreach ($handlers as $handler) {
        $this->whoops->pushHandler($handler);
    }
}
```

**Conclusion**

In my opinion handlers should be executed in the order they are registered but that would mean a breaking change.

Wether a handler is the last one executed or not should be a decision from the developer not from the handle but changing this to `pushHandler($handler, $last = false)` would also mean a breaking change.

The simplest solution is to allow the developer to "append" a handler by giving him a `unshiftHandler($handler)` method (to keep array method naming).

In case you don't have time to develop this but would accept such a PR I can prepare one for you.

denis-sokolov:
I agree that the handlers should be executed in the order they are registered. I would accept a PR which adds a proper `appendHandler` function and marks `pushHandler` as deprecated.

PR #630, opened 2019-06-22
feat!: add append and prepend handler instead of reverse execution

tflori:
see #580 

BREAKING CHANGE: since `pushHandler()` was actually prepending handlers
it is now an alias for prependHandler but `popHandler()` is still
removing the last handler.

This should make the interface clearer: pushing (append) is now executed at the end while prepend is alias by the deprecated `pushHandler()` method.

People using the deprecated `pushHandler()` method and `popHandler()` should update the implementation to use `shiftHandler()` and `prependHandler()`.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -29,7 +29,7 @@ final class Run implements RunInterface
     /**
      * @var HandlerInterface[]
      */
-    private $handlerStack = [];
+    private $handlerQueue = [];
 
     private $silencedPatterns = [];
 
@@ -41,7 +41,7 @@ final class Run implements RunInterface
     }
 
     /**
-     * Prepends a handler to the start of the stack
+     * Prepends a handler to the start of the queue
      *
      * @throws InvalidArgumentException  If argument is not callable or instance of HandlerInterface
      * @param  Callable|HandlerInterface $handler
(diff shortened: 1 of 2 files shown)


staabm:
IMO we should be very careful, because pushHandler() is one of the most important whoops APIs

tflori:
@staabm I guess that is the reason @denis-sokolov suggested to create new methods and mark this as deprecated.

Because of the reverse execution push was not actually appending handlers as the name indicates. The API is not changing but the naming to clear this out.

tflori:
sure thing

denis-sokolov:
🚀

filips123:
`pushHandler()` should probably also trigger `E_USER_DEPRECATED` to inform users who still use it.

GrahamCampbell:
Adding methods to an interface is a major breaking change.

filips123:
Ok, so it should be added in Whoops 3.0.
What about triggering `E_USER_DEPRECATED` on `pushHandler()`.

denis-sokolov:
The method will always work, there is no need to force users off it.

filips123:
But it is deprecated, so it could maybe be removed in some future major version.

icanhazstring:
Nice change 👍
I am fine with the depreciation warning. As in dev environment those warnings should pop up but not in production of course.

As an example: Zend did the same with the `getServiceLocator` method on controllers from v2 to v3. So this practice isn't new. So when it's deprecated, notify users about it using the warning and the `@deprecated` warning.

denis-sokolov:
Beware that these changes have been partially reverted in 2.6.0, we should not have released them in a patch version. Please note the PR and the accomodating work was great, the fault is thoroughly on me. https://github.com/filp/whoops/issues/639#issuecomment-568891526

Issue #535, opened 2017-10-04
Add Constants for editors


ximex:
I think it would be an improvement if there were constants for the supported editors

staabm:
what do you mean? what would this constant provided? when do you need those? which problem should they solve?

ximex:
It's not a problem. I think it's only more clear.
Code example:
```php
class PrettyPageHandler extends Handler
{
    const EDITOR_SUBLIME = "sublime";
    // ...
}

// somewhere in code
$handler->setEditor(PrettyPageHandler::EDITOR_SUBLIME);
```

denis-sokolov:
This is a good suggestion. If anyone is up to implement it, I would be open to accept a PR. For now, I will be closing this issue.

We keep the list of the most interesting features to add in our Wiki:
https://github.com/filp/whoops/wiki/Possible-features-to-add

PR #631, opened 2019-06-24
add editor-constants

ximex:
fixes: https://github.com/filp/whoops/issues/535

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -17,6 +17,17 @@ use Whoops\Util\TemplateHelper;
 
 class PrettyPageHandler extends Handler
 {
+    const EDITOR_SUBLIME = "sublime";
+    const EDITOR_TEXTMATE = "textmate";
+    const EDITOR_EMACS = "emacs";
+    const EDITOR_MACVIM = "macvim";
+    const EDITOR_PHPSTORM = "phpstorm";
+    const EDITOR_IDEA = "idea";
+    const EDITOR_VSCODE = "vscode";
+    const EDITOR_ATOM = "atom";
+    const EDITOR_ESPRESSO = "espresso";
+    const EDITOR_XDEBUG = "xdebug";
+
     /**
      * Search paths to be scanned for resources, in the reverse
      * order they're declared.


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -17,6 +17,17 @@
 
 class PrettyPageHandler extends Handler
 {
+    const EDITOR_SUBLIME = "sublime";
```
staabm: maybe we can add an inline comment with a link to https://github.com/filp/whoops/blob/master/docs/Open%20Files%20In%20An%20Editor.md ?

SvenRtbg: I would expect these constants to appear somewhere else. And there is a place to put them: `$this->editors` is the array defining return strings for 9 of these 10 constants.

Why are there 10 constants, and why is the `setEditor` method throwing an exception on anything that isn't either defined inside that array or otherwise a callable? Seems like one of these constants will probably not work right now.

SvenRtbg: Answer: xdebug is added dynamically based on the extension being installed or not. So xdebug isn't reliably available and will throw exceptions on systems where the extension is missing if you run `setEditor('xdebug')` (or use the new constant).

SvenRtbg:
On a personal note: I have mixed feelings about that editor feature as it currently is. You either provide a string that exists in the `$this->editors` array, or a callable.

But the lookup array isn't exactly a callable lookup for string arguments - it replaces itself with another string.

In order to add unknown editors that simply require a formatted string, one has to first call `addEditor` to inject an entry into that array. Then call `setEditor` with the new array key to activate.

Alternatively call `addEditor` with a callable and call `setEditor` with the key.

Alternatively call `setEditor` directly with the callable.

Note that any existing entry in the array will be overwritten, so you could still end up calling `setEditor('textmate')` and not getting the default Textmate config.

I think I'd prefer if `setEditor` would accept either a string that is listed in the preconfigured array, or that is treated like the string from the lookup array (setting an editor anonymously without calling `addEditor` before), and both cases will immediately convert it into a default callable. This would allow to decrease the cyclomatic complexity of the `getEditor` method (currently around 10, if I counted correctly).

PR #632, opened 2019-06-24
Make sure we close all output-buffer we started beforehand

staabm:


staabm:
hmm not sure I understand what the builds errors try to tell me.

staabm:
since the build errors are still there, I dont think this is the right fix for the issue at least.

maybe it fixes your problem, but at least not the one reported by travis ;-)

Magentron:
I'll try to have a better look at it later.

staabm:
Hhmm I guess the source of the travis build warnings is here:
https://github.com/filp/whoops/blob/3ab01b90babe76b4b8271da2e964056091a960a1/src/Whoops/Run.php#L345

Magentron:
FWIW, work-around for a test:

- before doing something:
```
        $level = ob_get_level();
```
- at the end of the test:
```
        while (ob_get_level() > $level) {
            $content = ob_get_clean();
        }
```

Magentron:
Who should I contact to get this PR merged?

denis-sokolov:
Me! Thanks for the notice, I’ll take a look at it eventually.

src/Whoops/Run.php
```diff
@@ -303,37 +303,39 @@ public function handleException($exception)
...
-            if (in_array($handlerResponse, [Handler::LAST_HANDLER, Handler::QUIT])) {
-                // The Handler has handled the exception in some way, and
-                // wishes to quit execution (Handler::QUIT), or skip any
-                // other handlers (Handler::LAST_HANDLER). If $this->allowQuit
-                // is false, Handler::QUIT behaves like Handler::LAST_HANDLER
-                break;
+        
+        try {
+            // Just in case there are no handlers:
+            $handlerResponse = null;
```
staabm: These assignments could be moved before the try/catch

Magentron: You are so right, my bad, preparing to update PR

denis-sokolov:
Thanks! 2.4.1.

PR #635, opened 2019-07-24
Change deprecated pushHandler to prependHandler

beebus:
On https://github.com/filp/whoops/blob/master/src/Whoops/Run.php on lines 43-54, it states that pushHandler is being replaced with prependHandler.

Specifically:
```
* @deprecated use appendHandler and prependHandler instead
     */
    public function pushHandler($handler)
    {
        return $this->prependHandler($handler);
    }
```


denis-sokolov:
Thank you for your contributions!

Issue #590, opened 2018-08-08
Configure editor to use xdebug automatically


tobiastom:
Currently one has to configure the opening of files inside an editor [manually](https://github.com/filp/whoops/blob/6b3abfd1a21cb180013e373b9b7ee5b64ffc6c3b/docs/Open%20Files%20In%20An%20Editor.md).

As the support for  [`xdebug.file_link_format`](https://github.com/filp/whoops/blob/e79cd403fb77fc8963a99ecc30e80ddd885b3311/src/Whoops/Handler/PrettyPageHandler.php#L112) is already there, I would like to submit a pull request that enables Xdebug as default editor automatically. 

This would make sense in my eyes as I can use a software that already uses whoops. As most of them do not have an editor configured, no linking is done. If Xdebug would be detected (and configured) automatically, I think the experience for users of that software would be much more intentional. 

A user did already show their intent to use this by defining `xdebug.file_link_format` in the first place.

staabm:
Not sure what you meand.

Dont the lines at https://github.com/filp/whoops/blob/e79cd403fb77fc8963a99ecc30e80ddd885b3311/src/Whoops/Handler/PrettyPageHandler.php#L112 already enable the xdebug editor?

tobiastom:
Not based on what I understand. If this check is successful, xdebug will be available inside the possible editors. It's not enabled by default.

denis-sokolov:
This is a good suggestion. If anyone is up to implement it, I would be open to accept a PR.

We try to keep the outstanding issues under control, so I will be closing this, as until someone implements this, there is no action to be done.

PR #638, opened 2019-08-05
Implemented automatic editor configuration if Xdebug is available.

tobiastom:
This pull request adds the feature to configure Xdebug as default editor, if the `xdebug.file_link_format` option is set, and Xdebug is loaded.

This should not break any existing functionality as the editor can be easily overwritten with the `setEditor` method.

This was previously mentioned in #590.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -126,6 +126,9 @@ class PrettyPageHandler extends Handler
             $this->editors['xdebug'] = function ($file, $line) {
                 return str_replace(['%f', '%l'], [$file, $line], ini_get('xdebug.file_link_format'));
             };
+
+            // If xdebug is available, use it as default editor.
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -126,6 +126,9 @@ public function __construct()
             $this->editors['xdebug'] = function ($file, $line) {
                 return str_replace(['%f', '%l'], [$file, $line], ini_get('xdebug.file_link_format'));
             };
+
+            // If xdebug is available, use it as default editor.
+            $this->editor = $this->editors['xdebug'];
```
denis-sokolov: Can we use the method for consistency?

```suggestion
            $this->setEditor('xdebug');
```

tobiastom: Of course. That's obviously the better way. Changed. :)

denis-sokolov:

PR #640, opened 2019-08-27
Fix BC break introduced in 2.4.0 release

### LINKED ISSUE #639: BC Break in 2.4.0
I believe version 2.4.0 contains BC Break (because of https://github.com/filp/whoops/pull/630) in method:

https://github.com/filp/whoops/blob/cde50e6720a39fdacb240159d3eea6865d51fd96/src/Whoops/Run.php#L130-L133

Because internal storage has been changed from stack to queue (reversed order) this method return now reversed order of handler than in version prior to 2.4.0.

Code to reproduce:
```php
$run->pushHandler($handler1);
$run->pushHandler($handler2);

assertSame([$handler1, $handler2], $run->getHanlders());
```

Works on <2.4.0 and fails on 2.4.0+.

/cc @tflori

michalbundyra:
Reverts all changes made to the tests which changes the behaviour (see #630).
Fixed `appendHandler`/`prependHandler` methods to add desired feature and
keep Backward Compatibility.

Adds test to check if behaviour of new methods is right.

Fixes #639

-----
Changes to make in V3 - BC Break:
- `getHandlers` should return handlers in order they are gonna be executed.
- `popHandler` should return the last handler to execute.
- `shiftHandler` should return the first handler to execute.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -29,7 +29,7 @@ final class Run implements RunInterface
     /**
      * @var HandlerInterface[]
      */
-    private $handlerQueue = [];
+    private $handlerStack = [];
 
     private $silencedPatterns = [];
 
@@ -41,7 +41,7 @@ final class Run implements RunInterface
     }
 
     /**
-     * Prepends a handler to the start of the queue
+     * Pushes a handler to the end of the stack
      *
      * @throws InvalidArgumentException  If argument is not callable or instance of HandlerInterface
      * @param  Callable|HandlerInterface $handler
@@ -54,7 +54,7 @@ final class Run implements RunInterface
     }
 
     /**
-     * Appends a handler to the end of the queue
+     * Adds a handler to be executed as last.
      *
      * @throws InvalidArgumentException  If argument is not callable or instance of HandlerInterface
      * @param  Callable|HandlerInterface $handler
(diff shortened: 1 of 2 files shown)


webimpress:
@denis-sokolov 
Please have a look on this PR. I am reverting all changes made to the tests and adding tests for `appendHandler`/`prependHandler` in according to what we want to accomplish by using these methods.

⚠️  Because design decisions made from the very beginning of the library stack is quite confusing, and popHandler / shiftHandler works "opposite":
- `popHandler` gives the last element added to the stack, so executed as first.
- `shiftHandler` gives the first element added to the stack, so executed as last.

💡 As `shiftHandler` was added in PR #630 I would consider removing it, or changing to match the functionality of `popHandler` - so returns the handler which will be executed as first. And deprecate `popHandler`.

💡 I would consider also changing all method names:
- `appendHandler` -> `append`
- `prependHandler` -> `prepend`
...
(deprecate old and suggest new, so we can have pop/shift with desired behaviour).

💡 The other solution I can suggest is to extend constructor and provide the "handler container". By default it will be LIFO (stack, as it is currently), but if needed it can be changed to FIFO (queue).
Then we can keep current behaviour very easily, without any confusion, and add new behaviour as well, when FIFO is used.





The user shouldn't

GrahamCampbell:
What's the decision here, then?

denis-sokolov:
I appreciate this PR, thank you. I am sorry I did not get to merge it to provide direct attribution, but I went the route of reverting previous changes and reintroducing a subset of changes from scratch. This was a quality PR and I hope this does not dissuade you from contributing in the future!

PR #649, opened 2019-12-22
Fixed branch alias

GrahamCampbell:


denis-sokolov:
Thank you. Is there a particular benefit from the alias? Perhaps we can drop it?

staabm:
The alias allows consumerS of the package to define a dependency onto the unfinished next iteration.

denis-sokolov:
Right, thank you. But I don’t think anybody uses this feature, it was outdated for a while. And we still release soon enough. ¯\\_(ツ)_/¯

GrahamCampbell:
Composer is able to compensate for incorrect branch aliases. It may have interpreted it as 1.x dev.

People do use branch aliases, and packages should maintain a correct indication of what the master branch tracks. This is important for people who test their libraries using dev stability installs, for example. They will always get dev versions, ignoring tags.

GrahamCampbell:
> It may have interpreted it as 1.x dev.

I of course meant 2.x-dev. ;)

PR #657, opened 2020-01-14
Fix anonymous class dumping in PrettyPageHandler

takamichi:
Fix this warning not to occur when an anonymous class instance is provided.

```
PHP Warning: class_parents(): Class @anonymous does not exist and could not be loaded in {PROJECT_DIR}/vendor/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php on line 148
```

When anonymous class instance, a warning occurs because `$class = '@anonymous'`.


staabm:
Can this change be covered by a unit test?

takamichi:
@staabm  I have no idea where to write what unit tests.

denis-sokolov:
Thank you for this! The $obj definitely seems more right to use than $class. It’s challenging to add a good class here, it’s alright.

PR #662, opened 2020-04-27
Add method for adding custom javascript files

basteyy:
I add a method whichs allows adding javascript-files to the layout. It's realized exactly like the custom css method.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -51,6 +51,13 @@ class PrettyPageHandler extends Handler
     private $customCss = null;
 
     /**
+     * The name of the custom js file.
+     *
+     * @var string
+     */
+    private $customJs = null;
+
+    /**
      * @var array[]
      */
     private $extraTables = [];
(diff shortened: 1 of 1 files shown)


denis-sokolov:
I think this makes sense as long as it is evident, that we don’t support custom js tweaks and they might break with minor Whoops updates.

@staabm, @GrahamCampbell, do you have any comments on this?

staabm:
maybe it would make sense to ask why this is needed or what it will be used for (use-case)?

basteyy:
In my case I used this method to add a javascript-based contact form to output. The alternative way is to change the layout or add another handler, which would break the correct dom (not so sexy).

staabm:
This sounds like you are using whoops in a production environment?

basteyy:
I knew that was coming, but no it isn't. It's the environment for the beta testers. ;)

basteyy:
And perhaps there should be a method for appending html-code as well. I mean if you allow to inject custom css you should also allow to inject javascript and html into the layout. Otherwise a method could be implemented, which allows to change the layout itself. This would be better than creating another handler which does more or less the same like the PrettyPageHandler.

staabm:
The html resources can already be overridden via https://github.com/filp/whoops/blob/1fd7ccb7d87d38d7a40210e1e058cd282bf746f3/src/Whoops/Handler/PrettyPageHandler.php#L605

basteyy:
Yep that's correct. Didn't' recognize that.

PR #663, opened 2020-04-28
PHPDoc fixes

GrahamCampbell:
Short descriptions should be at most 80 characters (counting from the far left, so actually 74 max of text). There then should be a blank phpdoc line then optional 80 character wrapped additional descriptions, then param docs, then throws docs, then return docs.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -29,8 +29,9 @@ class PrettyPageHandler extends Handler
     const EDITOR_XDEBUG = "xdebug";
 
     /**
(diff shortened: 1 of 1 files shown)


src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -696,8 +735,10 @@ public function setApplicationRootPath($applicationRootPath)
     /**
      * blacklist a sensitive value within one of the superglobal arrays.
      *
-     * @param $superGlobalName string the name of the superglobal array, e.g. '_GET'
```
GrahamCampbell, line 699: there was an actual syntax error here, before!

denis-sokolov:
Thank you!

PR #672, opened 2020-10-09
Support PHP 8.0

GrahamCampbell:
Runs HHVM and PHP 5 tests on travis. Runs PHP 7 and 8 on GitHub Actions.

### DIFF
--- a/tests/Whoops/RunTest.php
+++ b/tests/Whoops/RunTest.php
@@ -93,13 +93,15 @@ class RunTest extends TestCase
     }
 
     /**
-     * @expectedException InvalidArgumentException
      * @covers Whoops\Run::pushHandler
      */
     public function testPushInvalidHandler()
     {
         $run = $this->getRunInstance();
-        $run->pushHandler($banana = 'actually turnip');
+
+        $this->expectExceptionOfType('InvalidArgumentException');
+
+        $run->pushHandler('actually turnip');
     }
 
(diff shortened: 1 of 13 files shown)


GrahamCampbell:
Hmm, latest PHP 7.3 seems to have a bug causing infinite recursion...

GrahamCampbell:
Tests passing now. Travis run: . Actions run: https://github.com/GrahamCampbell/whoops/actions/runs/298433086.

NB @filp Travis is not showing up in the status because you need to switch to travis-ci.com from .org. Github actions will only start to show up after this PR is merged.

tests/Whoops/RunTest.php
```diff
@@ -402,6 +406,7 @@ public function testSilenceErrorsInPaths()
 
     /**
      * @covers Whoops\Run::handleError
+     * @requires PHP < 8
```
staabm, line 409: Why does this not work on php8

GrahamCampbell: Because this is a different kind of error now.

GrahamCampbell: PHP does not permit silencing it.

denis-sokolov (approved):
Thanks so much, Graham! Good effort has been put in here.

GrahamCampbell:
🎉

Issue #673, opened 2020-10-19
Add Exception code to JsonResponseHandler by default


CodeBrauer:
Any reason why the JsonResponseHandler is missing the Exception code?
Is there an easy way to enable it?

https://www.php.net/manual/en/exception.getcode.php

denis-sokolov:
I suppose an oversight that nobody has ever noticed! A PR for this will be accepted. Seems to be a small change in our formatter:

https://github.com/filp/whoops/blob/fa50d9db1c0c2fae99cf988d27023effda5524a3/src/Whoops/Exception/Formatter.php#L21-L26

CodeBrauer:
Yeah works fine.

```php
require __DIR__ . '/../vendor/autoload.php';
$run     = new Whoops\Run;
$handler = new Whoops\Handler\JsonResponseHandler();
$run->pushHandler($handler);
$run->register();
throw new Exception("Error Processing Request", 1024);
```

returns:

```json
{
  "error": {
    "type": "Exception",
    "message": "Error Processing Request",
    "code": 1024,
    "file": "/Users/gw/dev/pr-tester/whoops/examples/json.php",
    "line": 21
  }
}
```

If no code is given, it is `int:0` by default

PR #674, opened 2020-10-19
Add Exception getCode to formatExceptionAsDataArray

CodeBrauer:
According to #673

### DIFF
--- a/src/Whoops/Exception/Formatter.php
+++ b/src/Whoops/Exception/Formatter.php
@@ -21,6 +21,7 @@ class Formatter
         $response = [
             'type'    => get_class($exception),
             'message' => $exception->getMessage(),
+            'code'    => $exception->getCode(),
             'file'    => $exception->getFile(),
             'line'    => $exception->getLine(),
         ];


CodeBrauer:
I have also thought about adding it in `formatExceptionPlain', but this could be a breaking change for some users should they parse the plaintext with the current version.

denis-sokolov:
Yeah, JSON by default is more extendable, good call with the plain version. Did you try this code yourself already, does it work for you? It’s okay if you didn’t, but want to know what to expect!

denis-sokolov:
Missed comment from the other thread: https://github.com/filp/whoops/issues/673#issuecomment-712698277

Thanks for your contribution! Will release in a moment.

CodeBrauer:
> Yeah, JSON by default is more extendable, good call with the plain version. Did you try this code yourself already, does it work for you? It’s okay if you didn’t, but want to know what to expect!

Yeah works fine.

```php
require __DIR__ . '/../vendor/autoload.php';
$run     = new Whoops\Run;
$handler = new Whoops\Handler\JsonResponseHandler();
$run->pushHandler($handler);
$run->register();
throw new Exception("Error Processing Request", 1024);
```

returns:

```json
{
  "error": {
    "type": "Exception",
    "message": "Error Processing Request",
    "code": 1024,
    "file": "/Users/gw/dev/pr-tester/whoops/examples/json.php",
    "line": 21
  }
}
```

If no code is given, it is `int:0` [by default](https://www.php.net/manual/en/exception.getcode.php#111430) 

_Comment was in wrong thread_

PR #676, opened 2020-10-28
[Whoops\Run] Improve PHPdocs

AmraniCh:
Hi! this is some PHPDoc improvements and fixes, I've also moved the private `getInspector()` method to the bottom and `$canThrowExceptions` property to the top with other class properties.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -45,6 +45,7 @@ final class Run implements RunInterface
 
     /**
      * @var array
+     * @psalm-var list<array{patterns: string, levels: int}>
      */
     private $silencedPatterns = [];
 
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -7,6 +7,7 @@
 namespace Whoops;
 
 use InvalidArgumentException;
+use Throwable;
 use Whoops\Exception\ErrorException;
 use Whoops\Exception\Inspector;
 use Whoops\Handler\CallbackHandler;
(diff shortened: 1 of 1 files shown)


src/Whoops/Run.php
```diff
@@ -31,17 +43,39 @@ final class Run implements RunInterface
      */
     private $handlerStack = [];
 
+    /**
+     * @var array
```
staabm, line 47:
```suggestion
     * @var array
     * @psalm-var list<array{patterns: string, levels: int}>
```

src/Whoops/Run.php
```diff
@@ -31,17 +43,39 @@ final class Run implements RunInterface
...
+     * In certain scenarios, like in shutdown handler, we can not throw exceptions.
+     *
+     * @var bool
+     */
+    private $canThrowExceptions = true;
+
+    /**
+     * Run constructor.
+     *
+     * @param SystemFacade|null $system
```
staabm, line 66: in case where the php-native type already covers everything, the phpdoc itself is unnecessary/useless.

phpdoc-based-types should only be used when they add a value on top of the php-native type

AmraniCh: > in case where the php-native type already covers everything, the phpdoc itself is unnecessary/useless.

Yes, you're right. The phpdoc still useless especially if you use the scalar type-hinting in PHP7.

> phpdoc-based-types should only be used when they add a value on top of the php-native type

Whoops already uses phpdoc, I'm not adding something new.

staabm: > Whoops already uses phpdoc, I'm not adding something new.

My comment was related to the line beeing changed here - not about other changes within this PR.

E.g. the phpdoc for this construct() here does not add value (in comparison to the php native types).

I dont say you shouldn‘t add phpdocs as a whole, but just for certain situations

AmraniCh: This makes sense.

AmraniCh: I've doc for  `$system` parameter.

AmraniCh: I've added a description for  `$system` parameter.

denis-sokolov:

PR #678, opened 2020-11-05
Fix copy button issue

AmraniCh:
Hi! this is a slight fix of the copy button.

Before : 

whoops-fix.

After : 

whoops-fix2


denis-sokolov:
Thanks!

PR #680, opened 2020-11-15
Fixed invalid codeblock syntax

GrahamCampbell:


denis-sokolov:
How did that even happen? Thanks! :-)

PR #685, opened 2021-01-21
Fix styles of copy button after hover

gharlan:
since https://github.com/filp/whoops/pull/579 the styles of the copy button are set via the class `rightButton` instead of the previous id `#copy-button`.
But on mouseleave the class is set to `clipboard` and the `rightButton` class is missing.

Original button:
Bildschirmfoto 2021-01-21 um 22 12 06

Wrong style after hovering:
Bildschirmfoto 2021-01-21 um 22 12 19


denis-sokolov:
Thanks!

PR #687, opened 2021-03-16
Builder-style handlers

gwillz:
All add/set methods now self-return so handlers can be configured inline.

Particularly troublesome for PrettyPageHandler. It seems JsonPageHandler was already doing this for it's methods (`setJsonApi()`, `addTraceToOutput()`).

Like this:
```php
$run->addHandler((new PrettyPageHandler)
   ->setEditor('vscode')
   ->setPageTitle('An error occurred!')
   ->hideSuperglobalKey('_SERVER', 'SECURITY_KEY'));
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,9 @@
             "role": "Developer"
         }
     ],
+    "scripts": {
+        "test": "phpunit --testdox tests"
+    },
     "require": {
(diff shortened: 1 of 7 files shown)


denis-sokolov:
Fantastic, thank you!

Issue #688, opened 2021-03-18
Decide which exit code to send


lode:
It would be nice if we could decide which exit code gets send to CLI, in the same way we can do that with HTTP response codes. Right now `$this->system->stopExecution(1);` is being called which hard codes this.

I'm willing to setup a PR for this if this is a desired direction.

denis-sokolov:
Sounds like a perfectly valid option to add! I think it would be close to [allowQuit option](https://github.com/filp/whoops/blob/6ecda5217bf048088b891f7403b262906be5a957/src/Whoops/Run.php#L212) we have on Run.

PR #689, opened 2021-03-19
Allow to customize the exit code send in cli context

lode:
Fixes https://github.com/filp/whoops/issues/688

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -39,6 +39,11 @@ final class Run implements RunInterface
     private $sendHttpCode    = 500;
 
     /**
+     * @var integer|false
+     */
+    private $sendExitCode    = 1;
+
+    /**
      * @var HandlerInterface[]
      */
     private $handlerStack = [];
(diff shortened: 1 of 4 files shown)


src/Whoops/Run.php
```diff
@@ -288,6 +293,31 @@ public function sendHttpCode($code = null)
...
+     */
+    public function sendExitCode($code = null)
+    {
+        if (func_num_args() == 0) {
+            return $this->sendExitCode;
+        }
+
+        if ($code < 0 || 255 <= $code) {
+            throw new InvalidArgumentException(
+                "Invalid status code '$code', must be between 0 and 254"
```
denis-sokolov, line 314: To match the conditional above:

```suggestion
                "Invalid status code '$code', must be between 1 and 255"
```

tests/Whoops/RunTest.php
```diff
@@ -503,4 +503,33 @@ public function testSendHttpCodeWrongCode()
...
+    }
+
+    /**
+     * @covers Whoops\Run::sendExitCode
+     */
+    public function testSendExitCodeWrongCode()
+    {
+        $this->expectExceptionOfType('InvalidArgumentException');
+
+        $this->getRunInstance()->sendExitCode(255);
```
denis-sokolov, line 533: How does this test pass? The condition allows 255 with less-or-equal operator.

denis-sokolov (changes requested):
Superbly clean, thank you!

Some clarification with the allowed range is needed, could you take a look and clarify what you meant? Thanks.

tests/Whoops/RunTest.php
lode: It works for me. The conditional is about when to throw, not when to allow.

src/Whoops/Run.php
denis-sokolov: I’ve misread the conditional, I’ve read it as `!(0 < $code && $code <= 255)`. Yeah, your reasoning is stellar, thanks for the patient answer!

tests/Whoops/RunTest.php
denis-sokolov: Indeed, I’ve misread it. Thanks!

lode:
Thanks a lot for the quick working along!

PR #692, opened 2021-03-30
PrettyPageHandler allow hiding super globals with non-string value

### LINKED ISSUE #690: Masking only works on strings.
Hello,

Masking within `PrettyPageHandler` only masks the contents if it is a string

```
if (isset($superGlobal[$key]) && is_string($superGlobal[$key])) {
```

We have model's stored into the session, which I want to hide.
I could make an own `CustomPrettyPageHandler`
But the method in `PrettyPageHandler` is private so I would need to copy & paste the whole contents, which makes no sense for me.

Possible solutions:

- Allow hiding non-string contents (Just place *** with maybe `gettype($superGlobal[$key])`)
- Make the `mask`-method protected, so we can do our own stuff on that.

What you think?

prdatur:
- Enhance PrettyPageHandler to allow hiding super globals with non-string value. (non-string values will have a fixed asterisk length of 3)
- Fix missing @throws annotation.
- Fix typo in doc-block
- Add new string/non-string behaviour to doc-block description.

Fixes #690

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -175,6 +175,8 @@ class PrettyPageHandler extends Handler
 
     /**
      * @return int|null
+     *
+     * @throws \Exception
      */
     public function handle()
     {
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thanks, this looks great!

PR #694, opened 2021-04-14
Avoid redirects away from error page

fredden:
When a `location` header is set but no suitable HTTP response code accompanies this (eg, 201, 202, 301, 302, 303, 307, 308), [PHP will send a response code of `302`](https://www.php.net/manual/function.header.php). This will override any (incompatible) response codes set with `http_response_code()`.

With Whoops handling an error after something has set a `location` header, the error page isn't visible in the browser because of this feature. This pull request fixes this masking by removing any `location` headers before setting a suitable HTTP response code.


denis-sokolov:

PR #701, opened 2021-08-06
Add types that appendHandler & prependHandler functions can take

fujiwaraizuho:
Thank you for maintaining the library!

I noticed that `Callable` cannot be passed as an argument to `appendHandler` and `prepandHandler` in `Run.php`

https://github.com/filp/whoops/blob/fdf92f03e150ed84d5967a833ae93abffac0315b/src/Whoops/Run.php#L74-L81
https://github.com/filp/whoops/blob/fdf92f03e150ed84d5967a833ae93abffac0315b/src/Whoops/Run.php#L87-L94

`pushHandler`, which recommends the use of these functions, allows you to pass a `Callable` as an argument

I apologize if this is intentional.

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -501,7 +501,7 @@ final class Run implements RunInterface
     /**
      * Resolves the giving handler.
      *
-     * @param Callable|HandlerInterface $handler
+     * @param callable|HandlerInterface $handler
      *
      * @return HandlerInterface
      *
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -74,7 +74,7 @@ final class Run implements RunInterface
     /**
      * Explicitly request your handler runs as the last of all currently registered handlers.
      *
-     * @param HandlerInterface $handler
+     * @param Callable|HandlerInterface $handler
      *
      * @return Run
      */
@@ -87,7 +87,7 @@ final class Run implements RunInterface
     /**
      * Explicitly request your handler runs as the first of all currently registered handlers.
      *
-     * @param HandlerInterface $handler
+     * @param Callable|HandlerInterface $handler
      *
      * @return Run
      */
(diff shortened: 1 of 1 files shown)


src/Whoops/Run.php
```diff
@@ -74,7 +74,7 @@ public function __construct(SystemFacade $system = null)
     /**
      * Explicitly request your handler runs as the last of all currently registered handlers.
      *
-     * @param HandlerInterface $handler
+     * @param Callable|HandlerInterface $handler
```
staabm, line 77: you should also adjust the called `resolveHandler` methods phpdoc

fujiwaraizuho: Thank you!
I fixed it at [`59f635b` (#701)](https://github.com/filp/whoops/pull/701/commits/59f635b59788cab37a24c77b9a723d0cc9cc86a1)

staabm (approved):
are there other methods within this class which also accept the additional `Callable` and miss the phpdoc?

otherwise lgtm - thx for the contribution

fujiwaraizuho:
Probably not.
Thanks for the review!

src/Whoops/Run.php
zonuexe: @fujiwaraizuho Since [`callable`](https://www.php.net/manual/en/language.types.callable.php) is a type, not a class, it is usually written in all lowercase.

fujiwaraizuho: fixed! [`198e712` (#701)](https://github.com/filp/whoops/pull/701/commits/198e7120b4bb26204010ae80a11f9b77c641359d)

src/Whoops/Run.php
```diff
@@ -501,7 +501,7 @@ private function getInspector($exception)
     /**
      * Resolves the giving handler.
      *
-     * @param HandlerInterface $handler
+     * @param Callable|HandlerInterface $handler
```
staabm, line 504:
```suggestion
     * @param callable|HandlerInterface $handler
```

Issue #704, opened 2021-09-14
Occasional errors with zlib.output_compression = On


kexxt:
Hello, occasionally I experience this error when I should not be seeing any whoops-related pages. 

`Uncaught Whoops\Exception\ErrorException: ob_end_clean(): failed to discard buffer of zlib output compression (1)`

I've identified the issue is caused by `Run.php` line 396:

```
while ($this->system->getOutputBufferLevel() > 0) {
    $this->system->endOutputBuffering();
}
```

I am also submitting a PR with a very simple fix

kexxt:
Specifically the problem occurs when `ob_end_clean()` is called while `ob_get_length()` is zero. By checking for a length > 0, we avoid the aforementioned error. When I tested this fix I stopped seeing the error and was able to resume debugging my own application (I was actually supposed to see a whoops page for something different.

PR #705, opened 2021-09-14
Add output buffer length check before attempting to end it

ndcorder:
Hello, this PR is related to #704. Please let me know if any tests need to be adjusted.

Thank you!

### DIFF
--- a/src/Whoops/Run.php
+++ b/src/Whoops/Run.php
@@ -393,7 +393,7 @@ final class Run implements RunInterface
             // @todo Might be able to clean this up a bit better
             if ($willQuit) {
                 // Cleanup all other output buffers before sending our output:
-                while ($this->system->getOutputBufferLevel() > 0) {
+                while ($this->system->getOutputBufferLevel() > 0 && $this->system->getOutputBufferLength() > 0) {
                     $this->system->endOutputBuffering();
                 }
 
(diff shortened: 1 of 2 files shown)


staabm:
This fix looks bogus to me. I think you should investiage why/when this zlib error happens (e.g. in php-src)

kexxt:
Got it, I appreciate your input.

denis-sokolov:
I’m not sure it’s obviously bogus. We want to call ob_end_clean as long as there’s buffering present. ob_get_length and ob_get_level both seem like valid options to detect that, and if one’s more practical, we could use it.

It’d be nice to understand the root cause of it better. 🤔

staabm:
`ob_get_level` returns the number of nested output buffers (how many times ob_start() was called).
`ob_get_length` returns the number of bytes of the current active output buffer, no matter how nested the ob_start() calls are.

I think thats a totally different think

kexxt:
Basically I reach a point where `ob_get_length` returns 0 but `ob_get_level` returns 2. When it tries to `ob_get_clean`, zlib error occurs.

denis-sokolov:
@staabm, sure, they do different things, but they both are good candidates to use in that conditional. After all, is it correct to keep flushing _until we have no flushing enabled_ or _until we have nothing buffered_? Not obvious to me.

Issue #504, opened 2017-06-02
drop HHVM support


staabm:
HHVM is used nearly anywhere and since php7 is a thing also doesnt has a real use-case anymore.

it was a great project to boost zend-php though.

we should drop hhvm support and cleanup the code-base from hhvm workarounds (and drop hhvm ci).

do you guys agree?

jens1o:
I agree.

denis-sokolov:
I have fixed Travis builds ([now passing](https://travis-ci.org/filp/whoops/builds/238839186)).

This would technically make us bump a major version, so unless you show big improvements, I’d rather not.

prisis:
Only if we bump PHP to 7/7.1

denis-sokolov:
What do you mean “bump”?

prisis:
I mean dropping the supports for PHP 5 totally

denis-sokolov:
Thanks for the explanation. Yep, fully agree.

prisis:
Your welcome :)

staabm:
@denis-sokolov please add a milestone, so we dont miss this when we will create a new major version.
(we dont create a new major version for this topic but in the future we will have a new major and then this could/should not be forgotten)

HawkonDK:
I can help with that if you need help ;) know all the new tings in PHP 7.x

nrueckmann:
Don't know if this issue is still open (last comment is 2 years old) but AFAIK the common consensus is that updating the version of dependencies is not a BC break. It's more like ending support for this specific dependency as no public interface changes.

SvenRtbg:
There is this discussion https://github.com/semver/semver/issues/148#issuecomment-25529446

JefferyHus:
Isn't HHVM a runtime for Hack language? I don't believe dropping its support will be a good idea. For PHP version it's a must to stop supporting PHP 5.*

Rarst:
For trivia there are no longer any supported HHVM versions that support PHP code, the last one was [3.30 supported until November 2019](https://hhvm.com/blog/2018/12/17/hhvm-3.30.html). They went entirely Hack since, so there is nothing to support anymore.

GrahamCampbell:
Doesn't mean people aren't using it. PHP 5 is dead too...

GrahamCampbell:
and 7.0 and 7.1

rjcalifornia:
> Doesn't mean people aren't using it. PHP 5 is dead too...

[...]

PR #706, opened 2021-09-19
Add ReturnTypeWillChange to FrameCollection

kylekatarnls:
Fix PHP 8.1 compatibility

### DIFF
--- a/src/Whoops/Exception/FrameCollection.php
+++ b/src/Whoops/Exception/FrameCollection.php
@@ -10,6 +10,7 @@ use ArrayAccess;
 use ArrayIterator;
 use Countable;
 use IteratorAggregate;
+use ReturnTypeWillChange;
 use Serializable;
 use UnexpectedValueException;
 
@@ -89,6 +90,7 @@ class FrameCollection implements ArrayAccess, IteratorAggregate, Serializable, C
      * @see IteratorAggregate::getIterator
      * @return ArrayIterator
      */
+    #[ReturnTypeWillChange]
     public function getIterator()
     {
         return new ArrayIterator($this->frames);
@@ -98,6 +100,7 @@ class FrameCollection implements ArrayAccess, IteratorAggregate, Serializable, C
      * @see ArrayAccess::offsetExists
      * @param int $offset
      */
+    #[ReturnTypeWillChange]
     public function offsetExists($offset)
     {
         return isset($this->frames[$offset]);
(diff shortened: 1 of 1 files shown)


src/Whoops/Exception/FrameCollection.php
```diff
@@ -164,11 +172,22 @@ public function serialize()
...
 
+    public function __serialize(): array
+    {
+        return $this->frames;
+    }
+
+    public function __unserialize(array $serializedFrames): void
+    {
+        $this->frames = $serializedFrames;
+    }
```
denis-sokolov, line 189: Are these related to the ReturnTypeWillChange or is this an accidental addition?

denis-sokolov (commented):
Thank you, @kylekatarnls! Indeed, we want something like that.

src/Whoops/Exception/FrameCollection.php
kylekatarnls: This is also needed for PHP 8.1 compatibility because `Serializable` is deprecated.

denis-sokolov: Thanks!

SvenRtbg:
Guys, this merge destroys PHP 5 compatibility because of the return types.

kylekatarnls:
GitHub Actions started from PHP 7.0, I didn't notice the composer.json still support PHP 5.

If PHP 5 is still meant to be supported, the it should be added to GitHub Actions to be tested.

But maybe it's a mistake in the composer.json. Actually I don't think it will be possible to be compatible with both PHP 5 and 8.1 without deprecation/warning notices in the same branch.

And I know it's not what you want to hear but PHP 8 is out now, so it's really time to drop PHP 5.

If you decide not to upgrade PHP, then the more consistent move could be also to lock the `filp/whoops` version to something like `~2.13.0`

SvenRtbg:
I agree that the existing support for PHP 5, even though untested, is surprising, but it is what is announced in the composer.json.

As you can see from #504 it has been discussed long ago to drop PHP 5 support, but somehow never happened.

AmraniCh:
I think whoops must be able to run in all PHP versions (from PHP 5), PHP 5 is still there in the market and a lot of websites still use it specially in development environment (including me).

staabm:
just have a look at the master branch. denis already added testcoverage https://github.com/filp/whoops/commit/7bf80a3773ca927e3d55db2c07d72e55e295c9a9

kylekatarnls:
> PHP 5 is still there in the market and a lot of websites still use it specially in development environment (including me).

"A lot" is probably exaggerated:
https://blog.packagist.com/php-versions-stats-2021-1-edition/

It's completely negligible in proportion. And if you don't feel using PHP 5 is a problem, you should feel OK to not be up to date with the latest filp/whoops neither 🤷

PR #709, opened 2021-09-30
Update psr/log's version constraint.

ADmad:
With existing constraints it's not possible to install whoops in an app which has `psr/log` as a directly dependency with constraint `^3.0`.


denis-sokolov:

PR #719, opened 2022-02-12
Exclude `.github` folder from distribution package

pH-7:
Exclude `.github` folder as well from distribution package.


denis-sokolov:
Thank you!

PR #720, opened 2022-05-07
Update whoops.base.js

devilmengcry:
show scroll bar


denis-sokolov:
Thank you for your suggestion!

Currently the intent behind this piece of code is to show the preview of the source code. We’re only including more of it invisible for technical reasons, to enable syntax highlighting.

If we wanted to allow the user to scroll, it could be fun. But we’d need to include the entire source file then, and think through how the rest of the block works when you can scroll. If you think there’s a lot of value there, by all means, make a proposal and let’s discuss it!

PR #741, opened 2022-11-02
Upgrade prismJS due to a security issue

MrHHyde:
Upgrade the prism JS & CSS to the latest version due to an open security issue in the current version

Found in https://nvd.nist.gov/vuln/detail/CVE-2021-3801 
Patched in 1.29.0: https://github.com/PrismJS/prism/pull/3341


denis-sokolov:
Thanks, released 2.14.6!

Hopefully nobody is using Whoops in a context where this would be a vector of attack in any way.

PR #743, opened 2022-12-21
docs: added changes log for release version 2.14.6

AmraniCh:
As the title says, adding changes logs for version `2.14.6`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 2.14.6
+
+* Added support of PHP version `8.2` for GitHub workflow tests ([#740][i740]).
+* Upgraded prismJS to version `1.29.0` due to security issue ([#741][i741]).
+
+[i740]: https://github.com/filp/whoops/pull/740
+[i741]: https://github.com/filp/whoops/pull/741
+
 # 2.14.5
 
 * Allow ArrayAccess on super globals


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 2.14.6
+
+* Added support of PHP version `8.2` for GitHub workflow tests ([#740][i740]).
+* Upgraded prismJS to version `1.29.0` due to security issue ([#741][i741]).
+
+[i740]: https://github.com/filp/whoops/pull/740
+[i741]: https://github.com/filp/whoops/pull/741
```
denis-sokolov, line 7: I’ll remove the note of our internal work. It is not relevant to the consumers of the package, who read the changelog. Let me know if you mind it!

```suggestion
* Upgraded prismJS to version `1.29.0` due to security issue ([#741][i741]).

[i741]: https://github.com/filp/whoops/pull/741
```

denis-sokolov (commented):
Thank you!

CHANGELOG.md
AmraniCh: You're right it is relevant more to developers and maintainers, but in my opinion, It is a good practice to log everything either the internal relevant changes, I suggest to have like a standard for our change log file, for ex we can prefix the internal logs with `[Internal]: log_description`, so our two change logs will look like:

```markdown
* Added support of PHP version `8.2` for GitHub workflow tests ([#740][i740]).
* [Internal]: Upgraded prismJS to version `1.29.0` due to security issue ([#741][i741]).
```

Of course, all the internal changes should be added last and with each other, what do you think ?

denis-sokolov: I’d like to keep the main changelog focused on the information our consumers care about. If you want to maintain a separate internal changelog, I will merge that. Although evaluate whether the [list of commits](https://github.com/filp/whoops/commits/master) doesn’t solve that problem already.

denis-sokolov:
I’ve added your commit partially to the changelog (9c68d0d8b39). Thank you for correcting the oversight with the missed changelog entry!

CHANGELOG.md
AmraniCh: Okay, I will create a separate change log file (`CHANGELOG_INTERNAL.md`) for our internal relevant changes, I will maintain this file and watch the overall changes in Whoops and detect undocumented internal work changes.

AmraniCh: What do you think ? Or for things to be more organized we can wait until the next major update, then we start to log our internal changes.

denis-sokolov: I don’t think it‘s particularly valuable, but I also wouldn’t stop you if you think it is!

PR #748, opened 2023-01-25
Add insepctor interface + factory

gwillz:
This permits creation of custom inspectors within the default whoops runtime. Although inspectors are inheritable and usable with a custom runtime, this requires copy-pasting much of the runtime internals.

Primarily the goal of this is to provide one the ability to intercept traces - in particular trace rewrites for template languages (see [whoops-twig](//github.com/gwillz/whoops-twig)).

This isn't quite the same as #394 - although perhaps it might make a future revisit a bit easier.

### DIFF
--- a/src/Whoops/Exception/Inspector.php
+++ b/src/Whoops/Exception/Inspector.php
@@ -6,9 +6,11 @@
 
 namespace Whoops\Exception;
 
+use Whoops\Inspector\InspectorFactory;
+use Whoops\Inspector\InspectorInterface;
 use Whoops\Util\Misc;
 
-class Inspector
+class Inspector implements InspectorInterface
 {
     /**
      * @var \Throwable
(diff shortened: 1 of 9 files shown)


src/Whoops/Exception/Inspector.php
```diff
@@ -30,12 +32,19 @@ class Inspector
...
+     * @var \Whoops\Inspector\InspectorFactoryInterface|null
+     */
+    protected $inspectorFactory;
+
     /**
      * @param \Throwable $exception The exception to inspect
+     * @param \Whoops\Inspector\InspectorFactoryInterface $factory
      */
-    public function __construct($exception)
+    public function __construct($exception, $factory = null)
```
denis-sokolov, line 44: Do we have to allow this override? I’d prefer this if we can:

```suggestion
    public function __construct($exception)
```

denis-sokolov (commented):
👍

src/Whoops/Exception/Inspector.php
gwillz: Hey thanks for taking a look! Really appreciated.

I agree it feels a bit weird. I made it default null because I didn't want to break backwards compat should anyone be using the inspector outside of this project. It's also required because the instance needs to be retained for when new inspectors are created, such as `getPreviouslyExceptionInspector()`. 

That all said, there might be a better way? Maybe just a `setInspectorFactory()` method so we're not polluting the constructor.

Issue #550, opened 2017-11-22
Filter frames issue


DarthSaracen:
There are troubles when attempt to exclude some traces via filter() method in FrameCollection class.
private $frames' keys have gaps after array_filter() applied, so that may cause such errors as $frames->offsetGet(0)->isApplication() one, because there are no more frame with offset "0". I've temporary used that workaround `$frames->prependFrames([]);` after filter() applied, and it's working now, but I suppose that it's better to automatically reset the keys after filtering.

staabm:
please provide repro code to make your problem more clear.

DarthSaracen:
I'm using following code to filter "bad" frames:

`$handler->setInspector(new class ($exception) extends Inspector {

			public function getFrames() {
				$frames = parent::getFrames()->filter(function(Frame $error) {
					return
						$error->getClass() !== \Core\ErrorHandler::class
						&& strpos($error->getFile(), 'ErrorHandler') === false;
				});
                                // hack
				$frames->prependFrames([]);
				return $frames;
			}

		});`

SvenRtbg:
It would be more interesting to see your code that throws the error because of the missing index. I have the feeling that you should probably change this instead. 

Accessing a fixed array index that gets removed due to filtering does not sound like the best idea. Did you know that `reset()` not only rewinds an iterator, but also returns the first element?

DarthSaracen:
`$frames->offsetGet(0)->isApplication()` is not my code, it's from Whoops. By resetting keys I meant smth like array_map('array_values', $arr) that just reorder keys from 0 to N so indexes like [3, 4, 5] becomes [0, 1, 2]

SvenRtbg:
Ok, I think I get a better picture of what is going on:

The PrettyPageHandler here https://github.com/filp/whoops/blob/141494232ff7d5c565ce2a9cea2be0958243708c/src/Whoops/Handler/PrettyPageHandler.php#L218 assumes that the frames are zero-indexed, which is not always the case if you inject a filtering inspector object.

I'm not familiar with the internals, but my guess would be that Whoops should be using `reset()` to fetch the first frame instead of abusing `count()` (which returns the number of elements, not anything related to their indexes) and then blindly accessing the zero'th element.

denis-sokolov:
Thanks, @DarthSaracen! Fixed in 2.1.14.

Issue #410, opened 2016-04-25
How can I hide some trace frames?


neomerx:
Frames collections seems to have a method for filtering out some frames (e.g. closures and other non app specific ones). It's not clear how it should be configured

```php
$run     = new Run();
$handler = new PrettyPageHandler();

$appName = $config->getForApp(C::KEY_APP_NAME);

$handler->setPageTitle("Whoops! There was a problem with '$appName'.");

// I'd like to instruct handler to filter out some of the 'standard' frames for
// better readability. But how?

$run->pushHandler($handler);

$htmlMessage = $run->handleException($exception);
$response    = new HtmlResponse($htmlMessage, 500);
```

denis-sokolov:
I am not aware of any such functionality in Whoops. It would be pretty simple to add, though, PR is welcome!

WilliamStam:
>  It would be pretty simple to add

using eloquent as my DB handler, and the frames it returns are a mile long. looking to filter out the "vendor" frames.

would this be an option you guys would support?

```
$whoops->addTraceFilterCallable(funtion($frame){
    return true;
})
```

then `vendor/filp/whoops/src/Whoops/Exception/Inspector.php` the method getFrames() would read the function, apply the `array_filter` using the callable, ie `array_filter($this->addTraceFilterCallable,$this->frames)`

denis-sokolov:
@WilliamStam, that looks good. Maybe call it “addFrameFilter”, but we can bikeshed the PR later, of course.

WilliamStam:
@AmraniCh you are awesome! i was revisiting this again and started looking through the code to see if i can implement it.  

for anyone else finding this:

```
 $whoops = new \Whoops\Run;
 $whoops->addFrameFilter(function($item){
        if (!str_contains($item->getFile(), DIRECTORY_SEPARATOR . "vendor" . DIRECTORY_SEPARATOR)){
            return true;
        }
       return false;
 });
  
 $whoops->register();
```

AmraniCh:
@WilliamStam thanks, I believe your code example could be more simplified like this :

```php
$whoops = new \Whoops\Run;
$whoops->addFrameFilter(function($item) {
    return !str_contains($item->getFile(), DIRECTORY_SEPARATOR . "vendor" . DIRECTORY_SEPARATOR);
});
$whoops->register();
```

Or If you're running on PHP versions >= `7.4`, you can use arrow functions :

```php
$whoops = new \Whoops\Run;
[...]

PR #749, opened 2023-02-19
Support frame filters

AmraniCh:
As the title says, this PR concerns allowing filtering frames collected by the exception inspector (#410 - #550), this is I think a breaking change for users that have their own implementation of the run interface, if this PR is approved and merged, these users have to implement the new methods added to the interface (`addFrameFilter`, `getFrameFilters`, `clearFrameFilters`).


denis-sokolov:

PR #751, opened 2023-03-04
Fix the bug that call to a member function getFrameFilters() on null

huangdijia:
https://github.com/filp/whoops/blob/3e8aebbca9f0ae6f618962c4ad514077fd365ab3/src/Whoops/Handler/PrettyPageHandler.php#L307


limingxinleo:
+1

AmraniCh:
You're right, If we need to call a Whoops handler we have to set Its Run instance explicitly.

AmraniCh:
@denis-sokolov

denis-sokolov:
Included in 2.15.1.

Issue #747, opened 2022-12-25
Current line highlight does not work


denis-sokolov:
Since v2.14.6 the current line is not highlighted due to a bug upstream: https://github.com/PrismJS/prism/issues/3552.

denis-sokolov:
Fixed in Released in 2.15.2.

PR #756, opened 2023-04-12
fix(highlight): #747 Add line offset to highlight line

bart-jaskulski:
PrismJS sort of broken backward compatibility, while trying to improve highlight resilience[^1]. Since then it is required to add `data-line-offset` for any code, which doesn't start counting from 1, which is pretty each time, we deal with stack frames.

This commits add silently required data attribute in HTML, bringing back highlight feature and recovering from some fall through errors, raised by `TypeError` in JS execution flow (i.e. switching application and full frames tabs).

[^1]: https://github.com/PrismJS/prism/commit/9a4e725b76d22759fec58e1929450bbfe1bb6661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+* Fixed missing code highlight, which additionally led to issue with switching tabs, between application and all frames ([#747](https://github.com/filp/whoops/issues/747)).
+
 ## v2.15.1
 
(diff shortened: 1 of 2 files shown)


bart-jaskulski:
Actually, offset atrribute has started to be *required* since this commit: https://github.com/PrismJS/prism/commit/098e30008eefc45874f5a32b4ce7141534060890

denis-sokolov:
Thank you!

denis-sokolov:
Released in 2.15.2!

PR #758, opened 2023-07-12
Speed up syntax highlighting

garygreen:
This substantially reduces the time it takes to load a Whoops ever page, and time switching frame context.

### Error page with 96 frames, tested with Laravel
| Metric | Before PR | With PR | Improvement |
|---|---|---|---|
| Page load time | 1600ms | 900ms | +43% |
| Context switch time | 4000ms | 200ms | +95% |

This is because prior to this PR syntax highlighting occurs on page load, and every single click _for the whole page and all frames_, even those that aren't visible. This PR syntax highlights only frames that are visible.

😁

### DIFF
--- a/src/Whoops/Resources/js/whoops.base.js
+++ b/src/Whoops/Resources/js/whoops.base.js
@@ -25,7 +25,7 @@ Zepto(function($) {
    * highlight the current line
    */
   var renderCurrentCodeblock = function(id) {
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Thank you, this is an awesome thing to find! Also thank you for a clean PR with a good description!

PR #761, opened 2023-11-16
Correct vscode open links

andrewnicols:
The format of links for vscode is currently set to:

    vscode://file/{$file}:{$line}

This leads to URIs such as:

    vscode://file//Users/example/git/example.php

However vscode expects URIs in the format:

    vscode://file/Users/example/git/example.php

That is to say that the URI should be immediately after the `file`.

The correct URI format is:

    vscode://file{$file}:{$line}

See https://github.com/microsoft/vscode/issues/197319 for more information.

### DIFF
--- a/src/Whoops/Handler/PrettyPageHandler.php
+++ b/src/Whoops/Handler/PrettyPageHandler.php
@@ -118,7 +118,7 @@ class PrettyPageHandler extends Handler
         "macvim"   => "mvim://open/?url=file://%file&line=%line",
         "phpstorm" => "phpstorm://open?file=%file&line=%line",
         "idea"     => "idea://open?file=%file&line=%line",
-        "vscode"   => "vscode://file/%file:%line",
+        "vscode"   => "vscode://file%file:%line",
         "atom"     => "atom://core/open/file?filename=%file&line=%line",
         "espresso" => "x-espresso://open?filepath=%file&lines=%line",
         "netbeans" => "netbeans://open/?f=%file:%line",


denis-sokolov (commented):
Thank you for quickly fixing this.

I do agree with [your comment on the linked issue](https://github.com/microsoft/vscode/issues/197319#issuecomment-1813810445), I think this is a bug and a breking change in VS Code. The syntax looks wonky without the separator between the literal word “file” and the path to the file.

Let’s wait for a bit to see if that decision gets reverted. I have made [a PR](https://github.com/microsoft/vscode-docs/pull/6838) in an attempt to force the issue. If it ends up being merged, then that’ll be the official confirmation that the new syntax is canonical.

src/Whoops/Handler/PrettyPageHandler.php
```diff
@@ -118,7 +118,7 @@ class PrettyPageHandler extends Handler
         "macvim"   => "mvim://open/?url=file://%file&line=%line",
         "phpstorm" => "phpstorm://open?file=%file&line=%line",
         "idea"     => "idea://open?file=%file&line=%line",
-        "vscode"   => "vscode://file/%file:%line",
+        "vscode"   => "vscode://file%file:%line",
```
denis-sokolov, line 121: This seems possibly not correct. The new format is like this only on non-Windows. On Windows we still need a leading slash in the format. ([Possible update to VS Code docs](https://github.com/microsoft/vscode-docs/pull/6838/files))

We would need to introduce additional processing here: a conditional, or at least a str_replace. This is inconvenient. :-(

denis-sokolov:
It seems like this change is not needed after all: https://github.com/microsoft/vscode/issues/198509#issuecomment-1816205881.

Issue #764, opened 2024-09-03
PHP 8.4 Beta - Implicitly marking parameter $system as nullable is deprecated


3kbest:
Seen in: "filp/whoops": "2.15.4" and php:8.4.0beta4

PHP message: PHP Deprecated:  Whoops\Run::__construct(): Implicitly marking parameter $system as nullable is deprecated, the explicit nullable type must be used instead in /vendor-dev/filp/whoops/src/Whoops/Run.php on line 84

https://github.com/filp/whoops/blob/a139776fa3f5985a50b509f2a02ff0f709d2a546/src/Whoops/Run.php#L84

Can be fixed by adding ? before SystemFacade

PR #765, opened 2024-09-24
Add PHP 8.4 support

Slamdunk:
Fix https://github.com/filp/whoops/issues/764


Slamdunk:
So, the new syntax is incompatible with PHP versions prior to `7.1`, I had to drop them, and I think we can safely do that:

https://packagist.org/packages/filp/whoops/php-stats

image

It's time to say farewall to those 2.1% of developers that don't upgrade PHP since 2015 :vulcan_salute:

denis-sokolov:
Thank you! Released as 2.16.0.

PR #776, opened 2025-06-03
Prevent javascript error: `TypeError: can't access property "scrollIntoView", line is undefined`

staabm:
when navigating to a "[internal]" frame, no code block is rendered:

grafik

lets not assume in our javascript a code-block is always present.


staabm:
CI build fixed in https://github.com/filp/whoops/pull/777

PR #777, opened 2025-06-03
CI: use ubuntu 22.04

staabm:
Use slightly newer ubuntu version to make CI work again.
I don't not make it even newer, as I think we cannot support PHP 7.1 on newer ubuntu


denis-sokolov:
Thanks!

PR #778, opened 2025-06-11
Unregister/cleanup handler on destroy

pascalheidmann:
This PR ensures that the Run class unregisters itself if its destroyed (needed for longer running processes)


denis-sokolov:
Thank you! 2.18.2.

PR #780, opened 2025-06-24
Mention PHPStorm URL Handler for Linux

igorsantos07:
I found out via Perplexity why these handlers never really worked in my case (Linux Mint 21), even though it seemed to call PHPStorm from Vivaldi.

With that handler, it started (trying) to open the file on my IDE. Then, all that was left was write the custom handler to translate Docker paths into local ones (without the translating code, it just opens an empty file with the correct name lol)

Thanks for Whoops, sir!


denis-sokolov:
Thank you!

PR #781, opened 2025-08-04
Fix constant E_STRICT deprecation warning when running tests

TomA-R:
Fix constant E_STRICT deprecation warning when running tests on PHP >= 8.4:

```
$ vendor/bin/phpunit -c phpunit.xml.dist
PHP Deprecated:  Constant E_STRICT is deprecated in whoops/tests/bootstrap.php on line 8
```

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -3,8 +3,8 @@
  * Whoops - php errors for cool kids
  * @author Filipe Dobreira <http://github.com/filp>
  *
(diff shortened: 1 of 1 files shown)


denis-sokolov:
We [support PHP 7.1](https://github.com/filp/whoops/blob/5f9ed7ca9a65367f8c084dc069693a67a32f33d0/composer.json#L19), but for tests this seems to be a totally valid change, thanks!

